PR #7, opened 2017-09-04
Allow explicit band selection in both API and commandline tool

penntaylor:
This is a relatively simple change that allows the user to specify which bands to download. For users who need only a subset of the available bands, this reduces requirements for download time, network bandwidth, and storage space.

I apologize for the pull request with no prior contact about this feature. I implemented it for my own use and afterward realized it might be useful to others.


themre:
Thank you for your contribution!

PR #15, opened 2018-01-25
Guard against None or empty string data folders.

garbagemule:
This commit limits the possible values for the data_folder variables in the AwsService, DataRequest, and DownloadRequest classes to non-empty strings. This fixes the error thrown when providing no value for the data_folder argument in AwsProductRequest, and it enforces a consistent state across the classes where data_folder always holds a value that can be used with os.path.join.

Note that joining the empty string with another path is an idempotent operation, whereas joining with the current directory ('.') results in a new path that explicitly references "this directory". That is, joining the empty string with 'bob' results in 'bob', but joining the current directory with 'bob' results in './bob'. The two result paths are equivalent as long as any further path joining happens
via os.path.join (which it should).

### DIFF
--- a/sentinelhub/aws.py
+++ b/sentinelhub/aws.py
@@ -28,8 +28,8 @@ class AwsService(ABC):
                       If parameter is set to ``None`` the list will be set automatically.
     :type metafiles: list(str) or None
     """
-    def __init__(self, parent_folder='', bands=None, metafiles=None):
-        self.parent_folder = parent_folder
+    def __init__(self, parent_folder=None, bands=None, metafiles=None):
+        self.parent_folder = parent_folder or '.'
         self.bands = self.parse_bands(bands)
         self.metafiles = self.parse_metafiles(metafiles)
 
--- a/sentinelhub/data_request.py
+++ b/sentinelhub/data_request.py
@@ -32,7 +32,7 @@ class DataRequest(ABC):
     :type data_folder: str
     """
     def __init__(self, *, data_folder=None):
-        self.data_folder = data_folder.rstrip('/') if data_folder else None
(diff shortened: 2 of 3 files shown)


garbagemule:
I just noticed the Travis build failed. I tried reverting my changes and running the tests locally on the most recent commit before mine, but this fails as well (despite no failure in Travis on that commit). It looks like it's got something to do with a missing instance ID in the config-file.

Can you confirm whether or not this is actually an error on my part?

Thanks!

AleksMat:
A while ago me and my colleagues discussed if we want to set `data_folder` by default to `'.'` or to `None`. At the end we chose the option with `None`. The reason is that we wanted the user is aware of location where data is saved. That would prevent events when someone would accidentally save a lot of data to an unwanted location. Therefore it is required to define `data_folder` in case user wants to save data.

However in new version `1.0.5` we improved the error handling. Now user should get more descriptive error when trying to save data with unspecified `data_folder` parameter. 

About failed unit tests on Travis - no worries it was not your fault. There was a problem with build configuration. In the future merge requests won't be checked by Travis anymore since that requires an instance ID for Sentinel Hub services.

Thanks for helping us solve important package issues and sorry that we won't include this merge request.

garbagemule:
Ah yes, strong defaults vs. mandatory parameters is something that comes up quite a lot. I personally prefer strong defaults, but they are definitely not a silver bullet since, as you mention, they can lead to unexpected and unintended results. If the data folder is a mandatory parameter, it is definitely better to fail fast and explicitly than to let the program crash "at some point", so thanks for making it more explicit :)

And thanks again for the quick turnaround!

PR #23, opened 2018-03-01
Fixing sentinelhub imports in OGC web services example.

klemenkenda:


AleksMat:
Thanks for the fix, we somehow missed that.

PR #39, opened 2018-06-28
remove AWS credentials check and set in boto3 only if they are configured

chrischamberlin:
This enables retrieval of credentials from the environment as described at:
https://boto3.readthedocs.io/en/latest/guide/configuration.html

For example, boto3 will retrieve credentials from environment variables or the EC2 instance metadata service, allowing us to avoid hardcoding credentials in a configuration file.


AleksMat:
You are right, option to read credentials from the environment should be available.

Thanks for the contribution. The changes will be released today in package version `2.2.0`.

PR #53, opened 2018-11-27
Using geopedia filterExpression for querying

batic:
Added method to extend query sent to geopedia, allowing filtering directly on geopedia.

### DIFF
--- a/sentinelhub/geopedia.py
+++ b/sentinelhub/geopedia.py
@@ -248,3 +248,12 @@ class GeopediaFeatureIterator(GeopediaService):
         """
         for feature in self:
             yield feature['properties'].get(field, [])
+
+    def add_query(self, query=None):
+        """ Add query, e.g. f12345 = 2015
+
+        :param query: query string
+        :type query: str
+        """
+        if query:
+            self.query['filterExpression'] += ' && ' + query
--- a/sentinelhub/geopedia.py
+++ b/sentinelhub/geopedia.py
@@ -218,6 +218,7 @@ class GeopediaFeatureIterator(GeopediaService):
 
         self.gpd_session = GeopediaSession()
         self.features = []
+        self.layer_size = None
         self.index = 0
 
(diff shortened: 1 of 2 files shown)


sentinelhub/geopedia.py
```diff
@@ -248,3 +248,12 @@ def get_field_iterator(self, field):
...
             yield feature['properties'].get(field, [])
+
+    def add_query(self, query=None):
+        """ Add query, e.g. f12345 = 2015
+
+        :param query: query string
+        :type query: str
+        """
+        if query:
+            self.query['filterExpression'] += ' && ' + query
```
AleksMat: It is possible that `self.query` doesn't have `'filterExpression'` key. Therefore I suggest doing something like:
```Python
self.query['filterExpression'] = '{} && {}'.format(self.query['filterExpression'], query) if self.query.get('filterExpression') else query
```
And since you are repeatedly using string `'filterExpression'` I suggest declaring it in a constant inside the class.

sentinelhub/geopedia.py
```diff
@@ -248,3 +248,12 @@ def get_field_iterator(self, field):
         """
         for feature in self:
             yield feature['properties'].get(field, [])
+
```
AleksMat: Would it maybe make more sense to add this parameter to `__init__`? That is because once you start iterating and making HTTP requests it doesn't make much sense to change query parameter somewhere in the middle.

sentinelhub/geopedia.py
```diff
@@ -248,3 +248,12 @@ def get_field_iterator(self, field):
...
+    def add_query(self, query=None):
```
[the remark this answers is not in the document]
batic: Yes, having `query` as mandatory sounds better.

sentinelhub/geopedia.py
batic: It seems not possible to me, as `self.query['filterExpression']` is set in `__init__`. Otherwise, I agree.

PR #55, opened 2019-01-09
Update geo_utils.py

daviddemeij:
Add function to calculate the dimensions of a certain bounding box given the resolution. This is useful if you don't know the width and height of an image that you want to request.

This could maybe also be implemented in a WmsRequest to automatically determine width and height if these are not given, either by using a known resolution or by setting the resolution as an argument.

### DIFF
--- a/sentinelhub/geo_utils.py
+++ b/sentinelhub/geo_utils.py
@@ -12,6 +12,20 @@ LOGGER = logging.getLogger(__name__)
 
 ERR = 0.1
 
+def bbox_to_dimensions(bbox, resolution):
+    """ Calculates width and height in pixels for a given bbox of a given pixel resolution (in meters).
+    :param bbox: bounding box
+    :type bbox: common.BBox
+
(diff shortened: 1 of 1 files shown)


AleksMat:
Hi @daviddemeij,

thanks for the pull request. I agree with proposed new utility function. I made some minor changes to the code and added some unit tests. 

One thing I changed is that now width and height are being rounded to nearest integer instead of being cast down with `int` function. To me this makes a bit more sense, especially to avoid some precision errors. Is there any particular reason that you would wish to keep `int` instead of `round`?
I have made a PR to your fork so that you can review the changes: https://github.com/daviddemeij/sentinelhub-py/pull/1

daviddemeij:
Hey @AleksMat I agree with you that first rounding the numbers would make more sense, I was already thinkin about that.

AleksMat:
Excellent, you code has been merged to develop. It will be merged to master when we release a new version of the package, should be soon. Thanks. :+1:

PR #56, opened 2019-02-14
Feat/gpd session improve

AleksMat:
Changes:
- `GeopediaSession` class can now also handle login,
- `GeopediaSession` class now stores all information about current session,
- `GeopediaFeatureIterator` and `GeopediaImageRequest` can accept an instance of `GeopediaSession` as an optional parameter.

### DIFF
--- a/sentinelhub/geopedia.py
+++ b/sentinelhub/geopedia.py
@@ -62,7 +62,7 @@ class GeopediaSession(GeopediaService):
     :param password_md5: Password can optionally also be provided as already encoded md5 hexadecimal string
     :type password_md5: str or None
     :param is_global: If `True` this session will be shared among all instances of this class, otherwise it will be
-        used only with the single instance. Deafult is `False`.
+        used only with the single instance. Default is `False`.
     :type is_global: bool
     :param base_url: Base url of Geopedia REST services. If `None`, the url specified in the configuration
         file is taken.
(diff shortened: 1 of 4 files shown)


sentinelhub/geopedia.py
```diff
@@ -90,61 +101,87 @@ def session_id(self):
...
         it will also make login.
         """
-        self._session_end = datetime.datetime.now() + self.SESSION_DURATION
+        self._session_start = datetime.datetime.now()
 
         session_url = '{}data/v1/session/create?locale=en'.format(self.base_url)
-        self._session_id = get_json(session_url)['sessionId']
+        self._session_info = get_json(session_url)
 
         if self.username and self.password:
```
devisperessutti: I would add a check on whether either are specified, something like

```python
if self.username and self.password:
    self._make_login()
elif self.username or self.password:
    raise ValueError("Both username and password must be specified at login")
```

batic: Don't forget that user might have specified password_md5 (and not password)

AleksMat: Attributes `password` and `password_md5` are actually put into the same object `self.password` in `__init__`. At the moment if someone would for some reason specify both, `password` would have priority over `password_md5`.

I think current username/password handling is intuitive but just in case I can add in `__init__`:
- If someone specifies `password` and `password_md5` and error is raised.
- If only one of `self.username` and `self.password` are available an error is raised.

AleksMat: Added in https://github.com/sentinel-hub/sentinelhub-py/pull/56/commits/b48e7703a157531fedbaf8a6637c8e1a91b5b8d6

PR #64, opened 2019-04-19
uuid in request header

batic:
The uuid in header was used when debugging our requests on SH side.
The commit also includes a very simple exponential back-off.


AleksMat:
Closing this PR since it is quite outdated.

Regarding the proposed changes:
- Sending UUID for each request - I suggest re-thinking this if it is really needed. It can easily be added in the later versions.
- Exponential backoff - this is already implemented in version 3.0.
- Handling `requests.exceptions.ChunkedEncodingError` - this is a good idea and will be added today.

PR #70, opened 2019-05-04
Fix tile_id folder name

rouault:
The issue was seen on
https://github.com/OSGeo/gdal/issues/1507#issuecomment-489340153

The directory structure used to save the S2A_MSIL1C_20181119T031011_N0207_R075_T50TLK_20181119T061056
product is not consistent with the <IMAGE_FILE> values of the MTD_MSIL1C.xml file.
In that case, sentinelhub.aws creates GRANULE/L1C_T50TLK_A017804_20181119T031642
wheras it should be GRANULE/L1C_T50TLK_A017804_20181119T031012, as it is downloaded
from scihub.copernicus.eu


AleksMat:
Hi @rouault,

First of all thank you for letting us know about the problem and submitting this PR.

Unfortunately your solution is not yet good enough. The problem is that ESA has changed the naming structure of .SAFE format many times in last few years. To keep track of versions they put so called baseline number in product names (in your case that is N0207 -> baseline 02.07). So usually when they change something they bump the baseline number (but there are even cases where they didn't do that).

Last time we updated this functionality of `sentinelhub-py` was in August last year. At that point it supported baselines up to 02.06 for Sentinel-2 L1C products and 02.07 for Sentinel-2 L2A products. At the moment I see that the latest baselines are 02.07 for L1C and 02.12 for L2A.

It looks that up to some baseline last number in tile ID really was sensing time but now in 02.07 it is datastrip time. Therefore your change now creates wrong data structures for some of the previous baselines: 

```
Expected: S2A_MSIL1C_20170413T104021_N0204_R008_T31SCA_20170413T104021.SAFE/GRANULE/L1C_T31SCA_A009443_20170413T104021
Got: S2A_MSIL1C_20170413T104021_N0204_R008_T31SCA_20170413T104021.SAFE/GRANULE/L1C_T31SCA_A009443_0170610T103721

Expected: S2A_MSIL1C_20171012T112111_N0205_R037_T29SQC_20171012T112713.SAFE/GRANULE/L1C_T29SQC_A012046_20171012T112713
Got: S2A_MSIL1C_20171012T112111_N0205_R037_T29SQC_20171012T112713.SAFE/GRANULE/L1C_T29SQC_A012046_0171012T163950
```

We have [unit tests](https://github.com/sentinel-hub/sentinelhub-py/blob/master/tests/test_aws_safe.py#L51) for almost each supported baseline and with this change unit tests for L1C 02.04, 02.05 and L2A 02.05, 02.06, 02.07 don't work anymore. (I am not sure if baseline 02.06 is also wrong and we just have a bad example in tests.) By the way, only unit tests for L2A baselines require AWS credentials. So if you comment those out you can run L1C unit tests yourself without any cost.

So we would appreaciate a lot if you (or anyone else from community) could help us sort out the confusion with new baselines. Especially because this functionality is something that is mostly used by community and almost never internaly at Sinergise. Of course I will be happy to assist. :)

rouault:
I've updated my commit to be specific of L1C baseline 02.07 and extended test_aws_safe.py to cover this case

AleksMat:
Excellent, this seems to work better. However I found one problem:

When testing on a product `S2A_MSIL1C_20190129T143751_N0207_R096_T20LLK_20190225T132350`, its datastrip name was `S2A_OPER_MSI_L1C_DS_EPA__20190225T132350_S20190129T144524_N02.07` and your method extracted `0190225T132350` out of it instead of `20190129T144524`.

I slightly changed extraction method ([new code](https://github.com/sentinel-hub/sentinelhub-py/blob/rouault-fix_tileid_path/sentinelhub/aws_safe.py#L285)) and added 2 more unit tests.

Now I think the changes should be ready to merge to develop. I still don't know if current version now covers all possible cases. We will just have to see if we get some more bug reports in the future.

AleksMat:
Now I extended support for L2A baselines as well - currently should work up to 02.11. Also I added some warnings about new baselines in future.

The code has been merged to develop and will be today released in the new package version. Thank you @rouault for your help. :)

Issue #73, opened 2019-07-04
Segfault on import 


Thingus:
Hi maintainers

On AWS-hosted Ubuntu 18.04 server,

```python
import sentinelhub
```
fails with a segfault.
On strace, something seems to be repeatedly opening and closing the same file:
```
openat(AT_FDCWD, "/home/ubuntu/tools/miniconda/envs/pyeo_env/share/proj/epsg", O_RDONLY) = 6
close(6)
```
Using sentinelhub in a conda environment, installed with pip. Python 3.7.3, numpy 1.16.4 - everything else as installed in the pip. I'll put my whole environment in the first comment.

Thanks,

Thingus:
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                        main
blas                      1.0                         mkl
boto3                     1.9.35                    <pip>
botocore                  1.12.35                   <pip>
bzip2                     1.0.6             h14c3975_1002    conda-forge
ca-certificates           2019.6.16            hecc5488_0    conda-forge
cairo                     1.14.12              h8948797_3
certifi                   2019.6.16                py37_0    conda-forge
chardet                   3.0.4                     <pip>
Click                     7.0                       <pip>
curl                      7.61.0               h84994c4_0
Cython                    0.29                      <pip>
docutils                  0.14                      <pip>
expat                     2.2.6                he6710b0_0
fontconfig                2.13.0               h9420a91_0
freetype                  2.9.1                h8a8886c_1
freexl                    1.0.5                h14c3975_0
gdal                      2.3.0            py37h04863e7_1
geojson                   2.4.1                     <pip>
geomet                    0.2.0.post2               <pip>
geos                      3.6.2                heeff764_2
giflib                    5.1.4                h14c3975_1
glib                      2.56.2               hd408876_0
hdf4                      4.2.13               h3ca952b_2
hdf5                      1.8.18               h6792536_1
html2text                 2018.1.9                  <pip>
icu                       58.2                 h9c2bf20_1
idna                      2.7                       <pip>
intel-openmp              2019.0                      118
jmespath                  0.9.3                     <pip>
joblib                    0.12.5                   py37_0
jpeg                      9b                   h024ee3a_2
json-c                    0.13.1               h1bed415_0
kealib                    1.4.7                h5472223_5
krb5                      1.16.1               hc83ff2d_6
libblas                   3.8.0                    10_mkl    conda-forge
libboost                  1.67.0               h46d08c1_4
[...]

PR #77, opened 2019-07-29
Bugfix/prepare 2.6.0

AleksMat:
This branch contains a few fixes before the release:
- Problem with new versions of `pyproj`: https://github.com/sentinel-hub/sentinelhub-py/issues/73
- Renamed module `test_utils` to `testing_utils` (because of pytest)
- Fixed code coverage calculation and added support for codecov.

After this the package should be ready for release.


sentinelhub/constants.py
```diff
@@ -343,6 +343,9 @@ def projection(self):
         :return: pyproj projection class
         :rtype: pyproj.Proj
         """
+        if pyproj.__version__ >= '2':
+            pyproj.Proj(self.epsg, preserve_units=True)
+
```
iovsn: Return missing.

AleksMat: Nice catch. Unfortunately the problem goes deeper - it looks like this object cannot be initialized like this, otherwise it doesn't work correctly. I will investigate further.

AleksMat: I decided to revert this back to the way it was. Further explanation is here: https://github.com/sentinel-hub/sentinelhub-py/issues/73

PR #82, opened 2019-08-27
WIP: Adding support for BYOC.

gobaRules:
Support for BYOC added.

Tests and examples are WIP. We need to decide:
1) if and what data to use,
2) where to add the configuration for the layer, which works with BYOC. To able to run the current test, one needs to add a layer to her "Python" configuration as explained in JN with examples.


sentinelhub/constants.py
```diff
@@ -111,7 +112,41 @@ class _OrbitDirection(Enum):
...
 
 
-class DataSource(Enum):
+class DataSourceMeta(EnumMeta):
+    """
+    Meta class used to add custom (BYOC) datasources to DataSource enumerator.
+
+    BYOC stands for Bring Your Own Data. For more details see: https://docs.sentinel-hub.com/api/latest/#/API/byoc
+    """
+    def __call__(cls, collection_id, *args, **kwargs):
```
AleksMat: Here you assume that `collection_id` will always be string. Because of that things like `DataSource(DataSource.SENTINEL2_L1C)` or `DataSource(DataSource.SENTINEL2_L1C.value)` don't work anymore.

I recommend adding a check at the beginning of this method
```Python
if not isinstance(collection_id, str):
    return super().__call__(collection_id, *args, **kwargs)
```

sentinelhub/constants.py
```diff
@@ -111,7 +112,41 @@ class _OrbitDirection(Enum):
...
+        """
+        This is executed whenever DataSource('something') is called.
+
+        The method raises a ValueError if the 'something' does not match the format expected for collection id.
+        """
+        collection_id_pattern = '.{8}-.{4}-.{4}-.{4}-.{12}'
+        if not re.compile(collection_id_pattern).match(collection_id):
+            raise ValueError("Given collection id does not match the expected format {}".format(collection_id_pattern))
+
+        def _custom_datasource_name(collection_id):
```
AleksMat: In general defining a function inside a method is not very nice code practice. I recommend moving this to a static method of either `DataSourceMeta` or `DataSource` class.

sentinelhub/constants.py
```diff
@@ -111,7 +112,41 @@ class _OrbitDirection(Enum):
...
+            Prepares a name for custom (BYOC) datasource, which is then added to DataSource enum.
+
+            :param: collection_id: Collection id of the BYOC, user's input.
+            :type: string
+            :return: Name for custom (BYOC) datasource.
+            :rtype: string
+            """
+            return 'BYOC_{}'.format(collection_id)
+
+        if cls._member_names_.__contains__(_custom_datasource_name(collection_id)):
```
AleksMat: This part could also be written like this:

```Python
datasource_name = _custom_datasource_name(collection_id)
if datasource_name not in cls._member_names_:
    extend_enum(cls, datasource_name, collection_id)

return super().__call__(collection_id, *args, **kwargs)
```

sentinelhub/constants.py
```diff
@@ -160,6 +196,10 @@ def get_wfs_typename(cls, data_source):
         :rtype: str
         """
         is_eocloud = SHConfig().is_eocloud_ogc_url()
+
+        if 'BYOC_' in str(data_source):
```
AleksMat: A bit more exact check would be `if data_source.name.startswith('BYOC_'):`

sentinelhub/constants.py
```diff
@@ -160,6 +196,10 @@ def get_wfs_typename(cls, data_source):
...
+            return 'DSS10-{}'.format(str(data_source.value))
```
AleksMat: Whenever you use `format` you don't have to cast arguments to a string because it is done automatically. Besides, `data_source.value` should already be a string.

sentinelhub/constants.py
```diff
@@ -245,10 +287,20 @@ def get_available_sources(cls):
...
+               [custom_datasource for custom_datasource in cls if "BYOC_" in custom_datasource.name]
+
+    @classmethod
+    def get_custom_datasources(cls):
+        """Returns the list of all custom (BYOC) datasources, which have ben added to the DataSource enumerator.
 
+        :return: List of custom (BYOC) datasources
+        :rtype: list(sentinelhub.DataSource)
+        """
+        return [datasource for datasource in cls if "BYOC_" in datasource.name]
```
AleksMat: There should be 2 blank lines between classes (PEP8)

sentinelhub/constants.py
```diff
@@ -245,10 +287,20 @@ def get_available_sources(cls):
             return [cls.SENTINEL2_L1C, cls.SENTINEL2_L2A, cls.SENTINEL2_L3B, cls.SENTINEL1_IW, cls.SENTINEL1_EW,
                     cls.SENTINEL1_EW_SH, cls.SENTINEL3, cls.SENTINEL5P, cls.LANDSAT5, cls.LANDSAT7, cls.LANDSAT8,
                     cls.LANDSAT8_L2A, cls.ENVISAT_MERIS]
+
         return [cls.SENTINEL2_L1C, cls.SENTINEL2_L2A, cls.SENTINEL1_IW, cls.SENTINEL1_EW, cls.SENTINEL1_EW_SH,
                 cls.SENTINEL1_IW_ASC, cls.SENTINEL1_EW_ASC, cls.SENTINEL1_EW_SH_ASC, cls.SENTINEL1_IW_DES,
-                cls.SENTINEL1_EW_DES, cls.SENTINEL1_EW_SH_DES, cls.DEM, cls.MODIS, cls.LANDSAT8]
+                cls.SENTINEL1_EW_DES, cls.SENTINEL1_EW_SH_DES, cls.DEM, cls.MODIS, cls.LANDSAT8] + \
+               [custom_datasource for custom_datasource in cls if "BYOC_" in custom_datasource.name]
```
AleksMat: - I see you are using check `"BYOC_" in custom_datasource.name` in at least 3 different places. Maybe it would be nicer to add a method to `DataSource` class named `is_byoc`. This way you don't have to implement this check every time.

- Since you already have `get_custom_datasources` method you could use it here.

requirements.txt
```diff
@@ -10,3 +10,4 @@ shapely
 boto3
 botocore
 pyproj
+aenum
```
AleksMat: Coincidentally I found out that if `aenum<=2.1.2` function `extend_enums` doesn't work correctly (at least for some versions). Therefore I recommend setting here a condition `aenum>=2.1.4`.

tests/test_constants.py
```diff
@@ -99,5 +99,19 @@ def test_request_type(self):
             self.fail("Couldn't instantiate enum")
 
 
+class TestDataSource(TestSentinelHub):
+    def test_adding_custom_datasource(self):
+        collectionid_datasourcename_wfsid = (
+            ('0000d273-7e89-4f00-971e-9025f89a0000', 'BYOC_0000d273-7e89-4f00-971e-9025f89a0000',
+             'DSS10-0000d273-7e89-4f00-971e-9025f89a0000'),
+        )
+        for collectionId, dataSourceName, wfsId in collectionid_datasourcename_wfsid:
```
AleksMat: Variables should be named in lowercase_with_underscores and not in camelCase notation.

examples/ogc_request.ipynb
```diff
@@ -1608,6 +1608,70 @@
...
+      "needs_background": "light"
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "bbox =  [13.82387,45.85221, 13.83313, 45.85901]\n",
+    "byoc_bbox = BBox(bbox=bbox, crs=CRS.WGS84)\n",
+    "\n",
+    "INSTANCE_ID=\"65df328e-37d9-47d9-82a4-44aa96c57020\"\n",
```
AleksMat: We'll have to check if an instance ID should even be provided here...

tests/test_constants.py
```diff
@@ -99,5 +99,19 @@ def test_request_type(self):
...
+            datasourcename_tested = DataSource(collectionId).name
+            wfsid_tested = DataSource.get_wfs_typename(DataSource(collectionId))
+            self.assertEqual(dataSourceName, datasourcename_tested, msg="Expected {}, got {}".
+                             format(dataSourceName, datasourcename_tested))
+            self.assertEqual(wfsId, wfsid_tested, msg="Expected {}, got {}".format(wfsId, wfsid_tested))
+
```
AleksMat: Here you could also test calling various methods of `DataSource` class for BYOC data source constant (e.g. `is_sentinel1`, `is_timeless`, ...).

iovsn:
[Support a custom DataSource for working with BYOC data](https://trello.com/c/Q2CtiHMc/26-support-a-custom-datasource-for-working-with-byoc-data)

PR #84, opened 2019-09-23
threads by default is #processes multiplied by 5

PeterFogh:
Number of threads started for a request is by default (None) number of processes multiplied by 5, see
https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor

### DIFF
--- a/sentinelhub/data_request.py
+++ b/sentinelhub/data_request.py
@@ -95,7 +95,7 @@ class DataRequest(ABC):
             ``data_filter=[0, 2, -1]`` the method will return only 1st, 3rd and last item. Default filter is `None`.
         :type data_filter: list(int) or None
         :param max_threads: number of threads to use when downloading data; default is ``max_threads=None`` which
-            by default uses the number of processors on the system
+            by default uses the number of processors on the system multiplied by 5.
         :type max_threads: int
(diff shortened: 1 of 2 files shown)


AleksMat:
Hi Peter,

A while ago I think I confused specifications of `ThreadPoolExecutor` and `ProcessPoolExecutor` and made this mistake. Thanks for the fix, merging to develop. :+1:

PeterFogh:
Hi @AleksMat, perhaps your sentinel-hub doc-strings just should ref to the `TheadPoolExecutor` docs, as the number of default threads changes between versions of the "concurrent futures" package, see https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor. 
> Changed in version 3.5: If max_workers is None or not given, it will default to the number of processors on the machine, multiplied by 5, assuming that ThreadPoolExecutor is often used to overlap I/O instead of CPU work and the number of workers should be higher than the number of workers for ProcessPoolExecutor.
...
> Changed in version 3.8: Default value of max_workers is changed to min(32, os.cpu_count() + 4). This default value preserves at least 5 workers for I/O bound tasks. It utilizes at most 32 CPU cores for CPU bound tasks which release the GIL. And it avoids using very large resources implicitly on many-core machines.

PR #85, opened 2019-10-05
Feat/fis post request

batic:
Use post requests for all FIS requests; solves the issue of (large) geometries resulting in too long URL for get request.


sentinelhub/fis.py
```diff
@@ -29,10 +28,27 @@ def get_request(self, request):
...
+        params = self._get_common_url_parameters(request)
+        post_data = {**params, **self._get_fis_parameters(request, geometry)}
+        # post data keys on SH are expected to be in lowercase:
+        post_data = {k.lower(): v for k, v in post_data.items()}
+
+        url = self.get_base_url(request)
+        authority = self.instance_id if hasattr(self, 'instance_id') else request.theme
+
+        headers = OgcConstants.HEADERS
+        headers['Content-Type'] = 'application/json'
```
AleksMat: These two lines modify a dictionary `OgcConstants.HEADERS`, which could be later used for something else. Instead I suggest doing:

```Python
headers = {'Content-Type': 'application/json', **OgcConstants.HEADERS}
```
This will keep `OgcConstants.HEADERS` unchanged.

sentinelhub/fis.py
```diff
@@ -29,10 +28,27 @@ def get_request(self, request):
...
+
+        return DownloadRequest(url=f'{url}/{authority}',
```
AleksMat: You are using f-strings, that is why Travis CI fails for Python 3.5.
https://travis-ci.org/sentinel-hub/sentinelhub-py/jobs/593895803#L443

sentinelhub/fis.py
batic: Fixed, thanks.

sentinelhub/fis.py
batic: Not the appropriate place for such discussion, but any particular reason we keep 3.5 compatibility?

Issue #93, opened 2020-01-08
get response data without saving to file


akenney:
Is it possible in the current implementation to get the response image in the requested format but without saving to a file?  For example, if I specify the image_format as TIFF_d32f, then I believe that the SentinelHub server returns an actual tiff file, and if I specify save_data then I can get the contents by reading them back out of the file that gets created, but if I just get_data and don't save_data, then all the metadata gets thrown away and I'm only given the numpy array of numerical values.  I would like to be able to get the raw image bytes, because what I want to do is modify the numerical values and _then_ save to a file, using the original metadata.  It looks like the parsing of the response happens in the same function that makes the request (execute_download_request), so I think currently there is no function I can call that will give me the whole response content?

I could make a pull request to add this ability, but I am not sure what would be the preferred way of doing it.  Options include
- Keep all existing function inputs/outputs the same, pull out most of the body of execute_download_request (everything except the last little bit that does the image decoding) into a separate function that would return the non-decoded response, and have execute_download_request call that function and then decode the response.
- Add an argument to execute_download_request, specifying whether to return the whole response body or just the numpy array of data.
- Add an argument to the functions that create the request, specifying on the request object whether to return the whole response body or just the numpy array of data.  (This option seems more complicated.)

akenney:
I made a pull request: https://github.com/sentinel-hub/sentinelhub-py/pull/95
Seems to do what I am looking for.  I didn't add any specific test for it; it passes the existing tests that I can run, there are four that I can't run because they need to know the associated instance ID.

AleksMat:
We have reviewed the proposed changes and agree with the additional functionality. This way users will have an option to get back encoded data and decide to use different libraries to decode it.

The changes have been merged into package version 3.0, which will be released today.

Thanks for contributing. :)

PR #95, opened 2020-01-19
allow returning binary data instead of type-specific decoded data

akenney:
https://github.com/sentinel-hub/sentinelhub-py/issues/93


AleksMat:
These changes have been ported to the package version 3.0 and merged with the following PR https://github.com/sentinel-hub/sentinelhub-py/pull/108.

This will be released in version 3.0 today. Thanks for your contribution. :+1:

PR #96, opened 2020-01-22
Minor typo fixes found with codespell

schwehr:
Used this by hand (not the `-w` option)

```
codespell --version
1.15.0
```


schwehr:
Test failure is because `ppa:jonathonf/backports` has moved on to bionic.  See #97 where there is a different failure

AleksMat:
Thanks for the fixes. This will be merged into master on Monday. 

And thanks for pointing out the `codespell` package. We'll try to use it from now on during development. :+1:

PR #104, opened 2020-02-13
Implementation of the simple rate limit

iovsn:
Co-authored-by: AleksMat <matej.aleksandrov@sinergise.com>

### DIFF
--- a/sentinelhub/config.py
+++ b/sentinelhub/config.py
@@ -58,7 +58,8 @@ class SHConfig:
             ('max_opensearch_records_per_query', 500),
             ('max_download_attempts', 4),
             ('download_sleep_time', 5),
-            ('download_timeout_seconds', 120)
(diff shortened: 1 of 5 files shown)


sentinelhub/config.py
```diff
@@ -58,7 +58,8 @@ class _SHConfig:
             ('max_opensearch_records_per_query', 500),
             ('max_download_attempts', 4),
             ('download_sleep_time', 5),
-            ('download_timeout_seconds', 120)
+            ('download_timeout_seconds', 120),
+            ('number_of_download_processes', 1)
```
AleksMat: Could you also add the parameter description in `SHConfig` class docstring? I imagine this parameter could easily be misinterpreted if we don't explain what it means and what it is used for.

iovsn: Done.

PR #120, opened 2020-04-10
Better Sentinel5P support

holonomicjl:
- Support for non-eocloud Sentinel5P configuration (https://creodias.sentinel-hub.com)

- MINQA parameter support for Sentinel5P data

I tested the features by hand.


batic:
Thanks for the PR! 
We will have a look in next few days.

PR #126, opened 2020-06-23
Feat/ogc deprecation

mlubej:
Update examples to use processing API


mlubej:
> I have added a few minor fixes. Now I think it should be ready to merge.

OK, I don't think I have anything else to add. I'll merge.

Issue #128, opened 2020-06-28
Enable low resolution data access


mlubej:
In the [docs](https://docs.sentinel-hub.com/api/latest/#/data/Sentinel-2-L1C?id=previewmode) page it's stated that by default only resolutions to 250m area available. Would it make sense to add `previewMode: EXTENDED_PREVIEW` to the `SentinelHubRequest`? I noticed that I need to make the change here, since it's not possible to do it in `eo-learn.

AleksMat:
This functionality is already possible with `SentinelHubRequest` class. You can pass `previewMode` to a parameter `other_args` of `input_data` method. More in [documentation](https://sentinelhub-py.readthedocs.io/en/latest/sentinelhub_request.html#sentinelhub.sentinelhub_request.SentinelHubRequest.input_data).

mlubej:
> This functionality is already possible with `SentinelHubRequest` class. You can pass `previewMode` to a parameter `other_args` of `input_data` method. More in [documentation](https://sentinelhub-py.readthedocs.io/en/latest/sentinelhub_request.html#sentinelhub.sentinelhub_request.SentinelHubRequest.input_data).

I'm not sure that's correct, or I might not understand it correctly. I was aware of this parameter, but it didn't work the way I wanted it to.

Example here:

a) the original request

```javascript
{
  "bounds":{
    "bbox":[
      46.16,
      -16.15,
      46.51,
      -15.58
    ],
    "properties":{
      "crs":"http://www.opengis.net/def/crs/EPSG/0/4326"
    }
  },
  "data":[
    {
      "dataFilter":{
        "maxCloudCoverage":100,
        "mosaickingOrder":"mostRecent",
        "timeRange":{
          "from":"2020-06-12T00:00:00Z",
          "to":"2020-06-13T23:59:59Z"
        }
      },
      "type":"S2L1C"
    }
  ]
}
```

If I add `other_args = {'previewMode': 'EXTENDED_PREVIEW'}` I get

```javascript
{
  "bounds":{
    "bbox":[
      46.16,
      -16.15,
      46.51,
      -15.58
    ],
    "properties":{
      "crs":"http://www.opengis.net/def/crs/EPSG/0/4326"
    }
  },
  "data":[
    {
      "dataFilter":{
        "maxCloudCoverage":100,
        "mosaickingOrder":"mostRecent",
        "timeRange":{
          "from":"2020-06-12T00:00:00Z",
[...]

PR #132, opened 2020-07-20
WIP: fix updating with other_args dict

mlubej:
Properly update the dictionary objects when the `other_args` param is used.

The new function recursively loops over the `other_args` and only adds non-dict like values to assigned keys, meaning that you can provide a nested dictionary which will be used to update the internal dictionary object.

### DIFF
--- a/sentinelhub/sentinelhub_request.py
+++ b/sentinelhub/sentinelhub_request.py
@@ -10,6 +10,14 @@ from .geometry import Geometry, BBox
 from .time_utils import parse_time_interval
 
 
+def update_other_args(d, u):
+    for k, v in u.items():
+        if isinstance(v, dict):
+            update_other_args(d[k], v)
+        else:
+            d[k] = v
+
+
 class SentinelHubRequest(DataRequest):
     """ Sentinel Hub API request class
     """
@@ -123,7 +131,7 @@ class SentinelHubRequest(DataRequest):
             input_data_object['dataFilter']['collectionId'] = data_source.value
 
         if other_args:
-            input_data_object.update(other_args)
+            update_other_args(input_data_object, other_args)
 
         return input_data_object
 
@@ -155,7 +163,7 @@ class SentinelHubRequest(DataRequest):
             request_body['output'] = request_output
 
         if other_args:
-            request_body.update(other_args)
+            update_other_args(request_body, other_args)
 
         return request_body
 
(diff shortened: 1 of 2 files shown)


mlubej:
Closes #128

sentinelhub/sentinelhub_request.py
```diff
@@ -10,6 +10,18 @@
...
+def _update_other_args(dict1, dict2):
+    """
+    Function for a recursive update of `dict1` with `dict2`. The function loops over the keys in `dict2` and
+    only the non-dict like values are assigned to the specified keys.
+    """
+    for key, value in dict2.items():
+        if isinstance(value, dict):
+            _update_other_args(dict1[key], value)
+        else:
+            dict1[key] = value
```
AleksMat, line 22: There are 2 cases where this would fail:

- `value` is a dictionary but `key` doesn't exist in `dict1`
- `value` is a dictionary but `dict1[key]` is not a dictionary

I think the first one could at some point actually happen. Therefore I propose changing L19 into

```Python
if isinstance(value, dict) and isinstance(dict1.get(key), dict):
```

mlubej: Probably the second case is irrelevant in this case, or at least I don't see any cases where this could happen in the scope of general usage.

Regarding the first case you are right, but I changed it in the following way, so that if a key is not present in `dict1`, it is added to `dict1` regardless of the fact if he value is a dictionary or not

```
    for key, value in dict2.items():
        if key in dict1.keys():
            if isinstance(value, dict):
                _update_other_args(dict1[key], value)
            else:
                dict1[key] = value
        else:
            dict1[key] = value
```

mlubej: This fixed the failing test, but there is one other test which fails already on develop, perhaps something changed in the mean time. I'll update the values.

AleksMat: Yes, please update values for that one.

Regarding the change in `_update_other_args`, if you join both `if` sentences in your proposed solution you get:

```Python
    for key, value in dict2.items():
         if isinstance(value, dict) and key in dict1:
             _update_other_args(dict1[key], value)
        else:
            dict1[key] = value
```

Then the only difference is in `isinstance(dict1.get(key), dict)` vs `key in dict1`. The first option also covers a case when `dict1[key]` exists but is not a dictionary.

mlubej: Ah.. sorry, I'm a bit slow today. Thanks!

mlubej: Fixed. I also added a new test which normally fails at such resolution if the preview mode is not correct.

PR #138, opened 2020-09-07
Data source improvements

AleksMat:
Redefines data source object and enables users to define their own data sources. It also adds a new example notebook about data sources.


sentinelhub/data_sources.py
```diff
@@ -0,0 +1,522 @@
...
+
+from aenum import extend_enum
+
+from .config import SHConfig
+from .constants import ServiceUrl
+from .exceptions import SHDeprecationWarning
+
+
+class _Source:
+    """ Types of satellite sources
```
batic, line 18: The docstring could be removed, or should just say "Sentinel Hub collections" (collections being the term SH team is using).

sentinelhub/data_sources.py
```diff
@@ -0,0 +1,522 @@
...
+    BATCH = 'BATCH'
+    LANDSAT5 = 'Landsat 5'
+    LANDSAT7 = 'Landsat 7'
+    SENTINEL3 = 'Sentinel-3'
+    SENTINEL5P = 'Sentinel-5P'
+    ENVISAT_MERIS = 'Envisat Meris'
+
+
+class _SensorType:
+    """ Types of satellite sensors
```
batic, line 35: Types is not a correct word, and also it is a bit of a conundrum; MSI is a (satellite/sensor) type, but IW and EW are swath modes. I do not know what would be the "common denominator" for these.

I'll try to find a better name, cause for the moment I can't think of one.

sentinelhub/data_sources.py
```diff
@@ -0,0 +1,522 @@
...
+    MSI = 'MSI'
+    IW = 'IW'
+    EW = 'EW'
+    OLCI = 'OLCI'
+    SLSTR = 'SLSTR'
+    TROPOMI = 'TROPOMI'
+
+
+class _ProcessingLevel:
+    """ Types of processing level
```
batic, line 47: Types of processing level -> Processing levels

sentinelhub/data_sources.py
```diff
@@ -0,0 +1,522 @@
...
+    """
+    # pylint: disable=invalid-name
+    L1B = 'L1B'
+    L1C = 'L1C'
+    L2A = 'L2A'
+    L3B = 'L3B'
+    L2 = 'L2'
+    GRD = 'GRD'
+    MCD43A4 = 'MCD43A4'
```
batic, line 56: MCD43A4 is not a processing level, but rather a MODIS data set.

sentinelhub/data_sources.py
```diff
@@ -0,0 +1,522 @@
...
+
+
+class _Polarization:
+    """ Types of SAR polarizations
```
batic, line 60: Types of SAR polarizations -> SAR polarisations

Based on [SH docs](https://docs.sentinel-hub.com/api/latest/data/sentinel-1-grd/#polarization) there are more polarisations as defined here.

sentinelhub/data_sources.py
```diff
@@ -0,0 +1,522 @@
...
+
+class _Resolution:
+    """ Types of product resolution (specific to Sentinel-1 sources)
+    """
+    MEDIUM = 'MEDIUM'
+    HIGH = 'HIGH'
+
+
+class OrbitDirection:
+    """ Types of orbit directions
```
batic, line 77: Just leave out "Types of"

sentinelhub/data_sources.py
```diff
@@ -0,0 +1,522 @@
...
+    """ Different collections of bands
+    """
+    SENTINEL2_L1C = ('B01', 'B02', 'B03', 'B04', 'B05', 'B06', 'B07', 'B08', 'B8A', 'B09', 'B10', 'B11', 'B12')
+    SENTINEL2_L2A = ('B01', 'B02', 'B03', 'B04', 'B05', 'B06', 'B07', 'B08', 'B8A', 'B09', 'B11', 'B12')
+    SENTINEL3_OLCI = tuple(f'B{str(index).zfill(2)}' for index in range(1, 22))
+    SENTINEL3_SLSTR = ('S1', 'S2', 'S3', 'S4', 'S5', 'S6', 'S7', 'S8', 'S9', 'F1', 'F2')
+    LANDSAT8 = ('B01', 'B02', 'B03', 'B04', 'B05', 'B06', 'B07', 'B08', 'B09', 'B10', 'B11')
+    DEM = ('DEM',)
+    MODIS = ('B01', 'B02', 'B03', 'B04', 'B05', 'B06', 'B07')
+    SENTINEL1_IW = ('VV', 'VH')
```
batic, line 94: There are more polarisations available for acquisition modes [docs](https://docs.sentinel-hub.com/api/latest/data/sentinel-1-grd/#acquisitionmode). How do these link to the available bands?

sentinelhub/data_sources.py
```diff
@@ -0,0 +1,522 @@
...
+class _DataSourceMeta(EnumMeta):
+    """ Meta class that builds DataSource class enums
+    """
+    def __call__(cls, value, *args, **kwargs):
+        """ This is executed whenever DataSource('something') is called
+
+        This implements a shortcut to create a BYOC data source by calling DataSource('<BYOC collection ID>')
+        """
+        # pylint: disable=signature-differs
+        if not isinstance(value, str):
```
batic, line 111: Why not just raise an exception when using `DataSource('<BYOC collection ID>')` with message that one has to use `DataSource.from_byoc(collection_id)`?

sentinelhub/data_sources.py
```diff
@@ -0,0 +1,522 @@
...
+class OrbitDirection:
+    """ Types of orbit directions
+    """
+    ASCENDING = 'ASCENDING'
+    DESCENDING = 'DESCENDING'
+    BOTH = 'BOTH'
+
+
+class _Bands:
+    """ Different collections of bands
```
batic, line 85: Why are only some bands listed here? E.g. for S2L1C, there are [several other bands available on SH](https://docs.sentinel-hub.com/api/latest/data/sentinel-2-l1c/#available-bands-and-data).

batic (commented):
WIP, will continue with review tomorrow.

sentinelhub/data_sources.py
```diff
@@ -0,0 +1,522 @@
...
+        processing_level=_ProcessingLevel.GRD
+    )
+    ENVISAT_MERIS = DataSourceDefinition(
+        wfs_id='ENV.TILE',
+        service_url=ServiceUrl.EOCLOUD,
+        source=_Source.ENVISAT_MERIS
+    )
+
+    @classmethod
+    def define(cls, name, *, api_id=None, wfs_id=None, service_url=None, source=None, sensor_type=None,
```
batic, line 304: There are some additional collection parameters, e.g. view for [S3 SLSTR](https://docs.sentinel-hub.com/api/latest/data/sentinel-3-slstr-l1b/#view). I understand these parameters are "filters", but so are other parameters, that have their own argument here. 

Should the `define` also have an additional `**kwargs` parameter? (Despite how opposed I am to `**kwargs`, in this particular instance it might make sense)

sentinelhub/data_sources.py
AleksMat: Just trying to make a smoother transition with a warning. This way people (including us) have more time to change this in their codebase and in the meantime the code will still work.

sentinelhub/data_sources.py
```diff
@@ -0,0 +1,522 @@
...
+    api_id: str = None
+    wfs_id: str = None
+    service_url: str = None
+    source: str = None
+    sensor_type: str = None
+    processing_level: str = None
+    polarization: str = None
+    resolution: str = None
+    orbit_direction: str = None
+    bands: Tuple[str, ...] = None
```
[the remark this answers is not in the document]
AleksMat: No ellipsis is mean to be given as parameters. This is just the way how to define a type for a "tuple containing an arbitrary number of strings". If you write just `Tuple[str]` then Python interprets this as "a tuple containing a single string".

sentinelhub/data_sources.py
AleksMat: Done

sentinelhub/data_sources.py
AleksMat: I removed MCD43A4 as we don't need it anywhere.

sentinelhub/data_sources.py
AleksMat: Added additional polarizations. We don't use them anywhere but we can list them just in case.

sentinelhub/data_sources.py
AleksMat: Done.

sentinelhub/data_sources.py
AleksMat: I would leave the other polarizations and bands for users to define themselves. Here is just wanted to define the most commonly used ones.

sentinelhub/data_sources.py
AleksMat: I wanted to leave the other "bands" for users to define themselves. They are either metadata bands or products (cloud masks).

sentinelhub/data_sources.py
AleksMat: So far we have only parameters that are actually used somewhere in the code. So even if users would be able to add their own parameters these parameters would have no effect in the current code.

sentinelhub/data_sources.py
AleksMat: When I change `DataSource` to `DataCollection` I renamed this to `_CollectionType`.

sentinelhub/data_sources.py
AleksMat: Good point. For Sentinel-1 I added here `C-SAR` and moved `IW` and `EW` to a new parameter swath mode. Now it makes much more sense.

sentinelhub/areas.py
```diff
@@ -340,42 +346,49 @@ def _recursive_split(self, bbox, zoom_level, column, row):
...
+        :type data_source: DataCollection
+        """
         super().__init__(shape_list, crs, **kwargs)
 
-        if data_source is DataSource.DEM:
+        data_collection = data_source or data_collection
+        if data_source is not None:
+            warnings.warn('Parameter data_source is deprecated, use data_collection instead',
+                          category=SHDeprecationWarning)
+        if data_collection is DataCollection.DEM:
```
batic, line 385: The user is notified that she should not be using `data_source`, but nothing prevents her from doing so. And if she does, the following code will not "automagically" convert her `data_source` to `data_collection`. So the warning is in my opinion misleading, as I'd expect everything to work.

sentinelhub/data_collections.py
```diff
@@ -0,0 +1,546 @@
...
+    SENTINEL1_EW_SH = ('HH',)
+    SENTINEL5P = ('AER_AI_340_380', 'AER_AI_354_388', 'CLOUD_BASE_HEIGHT', 'CLOUD_BASE_PRESSURE', 'CLOUD_FRACTION',
+                  'CLOUD_OPTICAL_THICKNESS', 'CLOUD_TOP_HEIGHT', 'CLOUD_TOP_PRESSURE', 'CO', 'HCHO', 'NO2', 'O3',
+                  'SO2', 'CH4')
+
+
+class _DataCollectionMeta(EnumMeta):
+    """ Meta class that builds DataCollection class enums
+    """
+    def __call__(cls, value, *args, **kwargs):
```
batic, line 117: This method made sense when we've only refactoring DataSource. But with the changes to DataCollection I don't see this method as needed.

sentinelhub/data_request.py
```diff
@@ -287,11 +288,17 @@ def __init__(self, layer, bbox, *, time='latest', service_type=None, data_source
         :type data_folder: str
         :param config: A custom instance of config class to override parameters from the saved configuration.
         :type config: SHConfig or None
+        :param data_source: A deprecated alternative to data_collection
+        :type data_source: DataCollection
         """
+        if data_source is not None:
+            warnings.warn('Parameter data_source is deprecated, use data_collection instead',
+                          category=SHDeprecationWarning)
```
batic, line 296: Same comment as above.

sentinelhub/areas.py
```diff
@@ -340,42 +346,49 @@ def _recursive_split(self, bbox, zoom_level, column, row):
...
+        :param reduce_bbox_sizes: If `True` it will reduce the sizes of bounding boxes so that they will tightly fit
+            the given area geometry from `shape_list`.
+        :type reduce_bbox_sizes: bool
+        :param data_source: A deprecated alternative of data_collection
+        :type data_source: DataCollection
+        """
         super().__init__(shape_list, crs, **kwargs)
 
-        if data_source is DataSource.DEM:
+        data_collection = data_source or data_collection
```
batic, line 381: What happens if user adds both? data_collection should be favoured, right?

sentinelhub/sentinelhub_request.py
```diff
@@ -2,31 +2,22 @@
...
-        else:
-            dict1[key] = value
-
-
 class SentinelHubRequest(DataRequest):
     """ Sentinel Hub API request class
     """
     def __init__(self, evalscript, input_data, responses, bbox=None, geometry=None, size=None, resolution=None,
-                 **kwargs):
+                 base_url=None, **kwargs):
```
batic, line 20: Wouldn't it be better that base_url is taken from DataCollection upon requesting data?

Or is there any other use-case, where duplicity is needed?

sentinelhub/sentinelhub_request.py
```diff
@@ -87,52 +84,61 @@ def create_request(self):
...
                            by it.
         :param other_args: dict
+        :param data_source: A deprecated alternative of data_collection
+        :type data_source: DataCollection
         """
+        data_collection = data_source or data_collection
+        if data_source is not None:
+            warnings.warn('Parameter data_source is deprecated, use data_collection instead',
+                          category=SHDeprecationWarning)
+        if not isinstance(data_collection, DataCollection):
```
batic, line 109: Why is not such check done also in other similar cases?

sentinelhub/data_collections.py
AleksMat: I agree. But I left it only for batchward-compatibility reasons. This way `DataSource('byoc-id')` still works but raises a deprecation warning.

sentinelhub/areas.py
AleksMat: Handled with `handle_deprecated_data_source`

sentinelhub/sentinelhub_request.py
AleksMat: I now added casting `DataCollection(data_collection)` everywhere where needed. This has more or less the same effect than the check above.

sentinelhub/exceptions.py
```diff
@@ -45,6 +45,28 @@ class SHRuntimeWarning(RuntimeWarning):
     """
 
 
+def handle_deprecated_data_source(data_collection, data_source, default=None):
```
batic, line 48: I like the approach, but I'd suggest putting this to data_collections.py.

sentinelhub/data_request.py
batic: I like the solution with `handle_deprecated_data_source` method. Closing.

sentinelhub/exceptions.py
AleksMat: You are right. I moved it to data_collections.

tests/test_geopedia.py
```diff
@@ -63,8 +63,8 @@ def test_return_type(self):
         self.assertEqual(len(self.data), data_len,
                          "Expected a list of length {}, got length {}".format(data_len, len(self.data)))
 
-    def test_stats(self):
-        self.test_numpy_data(np.array(self.data), exp_min=0, exp_max=255, exp_mean=150.9248, exp_median=255)
+    # def test_stats(self):
```
batic, line 66: Why is this commented out? If it is not to be used, just remove it.

batic:
I only have one last comment regarding commented out test. After that, the PR can be merged. (I might suggest squashing the commit, though).

tests/test_geopedia.py
AleksMat: This integration test sometimes fails if Geopedia is under heavy load (like yesterday). We have an issue to improve this but it is not a high priority. Commenting this out was not the best solution therefore I now updated it so that the test is marked with `@pytest.mark.xfail`

PR #142, opened 2020-09-17
Feat/config file update

### LINKED ISSUE #139: Better handling of config.json file
This was mentioned already several times, just wanted to put down a ticket so that we don't lose track of this.

The `config.json` template file for storing credentials is a part of the repo. When we download the repo and change this file, it tracks the changes. Instead of this, it would be preferred if the file is ignored by GIT so that I can focus on actual changes in the development process and so that I don't accidentally delete/overwrite my credentials when reverting the changes.

The `.json` format is in the `.gitignore`, but it's ignored because this file is already in the repo. Not sure what the best solution is. Perhaps an auto-generated template file when it's first called or something.

mlubej:
This PR removes the `config.json` from the repo. The default file is created if the config is called and if it doesn't exist.

Closes #139


sentinelhub/config.py
```diff
@@ -90,17 +90,18 @@ def _parse_configuration(self, config):
...
+        def get_config_file(self):
+            """ Checks if configuration file exists and returns its file path.
+            If file doesn't exist it creates a default configurations file.
 
             :return: location of configuration file
             :rtype: str
             """
             config_file = os.path.join(os.path.dirname(__file__), 'config.json')
 
             if not os.path.isfile(config_file):
```
batic, line 102: No need to change, but using `isfile` will fail in (rather improbable) case of having an existing directory named 'config.json'...

batic:
👍 from me.

sentinelhub/config.py
mlubej: It does seem rather improbable, but thanks for pointing it out.

AleksMat: I think that in such an improbable case it would be completely ok if it fails.

AleksMat (commented):
Looks ok. As far as I see there are a couple of use cases we would like to cover:

- When we release the package to PyPI we call `sentinelhub.config --clear` and then upload the package. So if `config.json` wouldn't exist then `sentinelhub.config --clear` would trigger its creation. Therefore `config.json` will always be released together with the package and anyone who istalls it with pip will have it.
- If a developer clones a git repo and install the package e.g. with `pip install -e .` then `config.json` won't be created automatically. But often during development it's convenient to have it so that you can just copy-paste credentials into it.

Because of the 2nd case I suggest adding the following function into `setup.py`

```Python
def try_create_config_file():
    """ After the package is installed it will try to trigger saving a config.json file
    """
    try:
        from sentinelhub.config import SHConfig
        SHConfig()
    except BaseException:
        pass
```
and calling it at the end after the `setup` is called. I checked that this then `config.json` automatically right after a package is installed.

mlubej:
>     * When we release the package to PyPI we call `sentinelhub.config --clear` and then upload the package. So if `config.json` wouldn't exist then `sentinelhub.config --clear` would trigger its creation. Therefore `config.json` will always be released together with the package and anyone who istalls it with pip will have it.

Do you mean `-- reset`? Otherwise, I agree.

> Because of the 2nd case I suggest adding the following function into `setup.py`
> 
> ```python
> def try_create_config_file():
>     """ After the package is installed it will try to trigger saving a config.json file
>     """
>     try:
>         from sentinelhub.config import SHConfig
>         SHConfig()
>     except BaseException:
>         pass
> ```
> 
> and calling it at the end after the `setup` function is called. I checked that this creates `config.json` automatically right after a package is installed.


Cool, I was trying to find an optimal way of doing this because I also recognized this issue, but didn't know how exactly to approach it. Will this cause any problems on Windows? Should it be run inside of `if __name__ == "__main__":` or something?

AleksMat:
> Do you mean `-- reset`? Otherwise, I agree.

Sorry, I did mean `--reset`.

> Cool, I was trying to find an optimal way of doing this because I also recognized this issue, but didn't know how exactly to approach it. Will this cause any problems on Windows? Should it be run inside of `if __name__ == "__main__":` or something?

I don't see any problems with Windows. Besides, everything is under `try-catch`. That is just in case if there exists another way of installing a package using `setup.py`, that we don't know about, which wouldn't work like that.

I think adding `if __name__ == "__main__":` would not be ok. That is because when you install it with pip from command line I think `setup.py` is not the main script.

mlubej:
OK cool, just wanted to check. Thanks! Will add this to the setup file.

AleksMat (approved):

PR #144, opened 2020-10-06
CRS parsing shows a warning instead of an error

AleksMat:
When parsing `sentinelhub.CRS(pyproj.CRS(4326))` we now get a warning instead of an explicit error about the coordinate order. This way it is up to a user to correctly deal with coordinates.


batic:
👍

PR #145, opened 2020-10-07
Moved get_json and get_xml functions to DownloadClient methods

AleksMat:
The following enables that `get_json` and `get_xml` helpers can be used by classes that inherit from `DownloadClient` without a need for another helper function.


batic:
To me this looks very nice now. 

It is completely transparent (to the user/programmer) which client is being used to get the data. 

👍

batic (approved):

PR #154, opened 2021-02-10
add floats to config allowed parameters

DomagojKorais:
`download_timeout_seconds` and `download_sleep_time` can now be float or int type, instead of only int.

### DIFF
--- a/sentinelhub/config.py
+++ b/sentinelhub/config.py
@@ -4,6 +4,7 @@ Module for managing configuration data from `config.json`
 
 import os
 import json
+import numbers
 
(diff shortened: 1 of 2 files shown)


AleksMat (commented):
Besides a small pylint issue looks ok. My suggestion would also be that a new unit test would be added [here](https://github.com/sentinel-hub/sentinelhub-py/blob/master/tests/test_config.py) that would check that saving an int or float works, but e.g. saving a string doesn't.

DomagojKorais:
added tests

PR #155, opened 2021-02-10
use parsing function when retriving type

DomagojKorais:
MimeType has an already implemented function to parse types, the function implement the logic to deal with parsing of  "jpeg"  and "jpg" file formats.

### DIFF
--- a/sentinelhub/sentinelhub_request.py
+++ b/sentinelhub/sentinelhub_request.py
@@ -43,7 +43,8 @@ class SentinelHubRequest(DataRequest):
         if not isinstance(evalscript, str):
             raise ValueError("'evalscript' should be a string")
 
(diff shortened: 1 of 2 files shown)


AleksMat (commented):
I recommend that one of the integration tests [here](https://github.com/sentinel-hub/sentinelhub-py/blob/master/tests/test_sentinelhub_request.py) requests a jpeg response. So there are 2 options:

- update one of the existing tests to use jpeg
- add a new test

In this case, I'm more inclined towards the 1st option since we don't want to have too many itegration tests with SH service. So if each test can be used for more than one purpose that is great.

DomagojKorais:
updated one of the existing tests in order to request a jpeg instead of a tif.

PR #156, opened 2021-02-10
improve error handling for batch

DomagojKorais:


sentinelhub/sentinelhub_batch.py
```diff
@@ -405,6 +405,9 @@ def _iter_pages(service_url, config, **params):
         results = client.get_json(url, use_session=True)
 
         results_data = results.get('data') or results.get('member')
+        if results_data is None:
+            raise Exception("No tiles found, please run analysis on batch request before calling this method.")
```
AleksMat, line 409: Let's raise a more specific exception. That way it will be easier to catch it.

At first, I was thinking about `RuntimeError` but perhaps it would be even better to define a new custom exception in [exceptions.py](https://github.com/sentinel-hub/sentinelhub-py/blob/master/sentinelhub/exceptions.py) that would inherit from `BaseSentinelHubException` and use that one?

DomagojKorais: Custom exception added.

AleksMat: Excellent, I agree with the change.

PR #158, opened 2021-02-11
Fixed integration tests

AleksMat:
All tests should now be up-to-date with recent changes in SH service and pass successfully.


batic:
Looks ok, tests now pass. I'd now concatenate some split lines (they are not long enough to merit newline split), but that does not impact the functionality whatsoever. 

A side note: did you visualise the various OGC requests to understand the different statistics? Any other difference apart from simple/V3 evalscript related transitions?

batic (approved):

AleksMat:
> A side note: did you visualise the various OGC requests to understand the different statistics? Any other difference apart from simple/V3 evalscript related transitions?

I didn't visualize but the main things I noticed are:

- changes are only for bands layer, true color layers remain unchanged,
- all bands layers now return dataMask as an additional band
- all bands layer now return data in FLOAT32 format
- Sentinel-1 and I think also Sentinel-3 layers return nan values for areas with no data FLOAT32 format. This is a bit inconsistent with e.g. Sentinel-2 and Landsat 8 which return zeros but in general not a problem.

PR #161, opened 2021-02-15
don't create bbox if not specified by user

DomagojKorais:
bbox is not created if the user doesn't specify it.
I modified an already existing test in order to avoid doing too many requests. The modified test is `test_multipart_geometry`, since the results doesn't change from the previous version, where bbox was specified explicitly, I would say that there is no reason to create the bbox.

### DIFF
--- a/sentinelhub/sentinelhub_request.py
+++ b/sentinelhub/sentinelhub_request.py
@@ -225,9 +225,6 @@ class SentinelHubRequest(DataRequest):
         if bbox and geometry and bbox.crs != geometry.crs:
             raise ValueError("bbox and geometry should be in the same CRS")
 
-        if bbox is None:
-            bbox = geometry.bbox
-
         crs = bbox.crs if bbox else geometry.crs
 
         request_bounds = {
--- a/tests/test_sentinelhub_request.py
+++ b/tests/test_sentinelhub_request.py
@@ -360,7 +360,6 @@ class TestSentinelHubRequest(TestSentinelHub):
                 SentinelHubRequest.output_response('default', MimeType.TIFF),
                 SentinelHubRequest.output_response('userdata', MimeType.JSON)
             ],
-            bbox=bbox,
             geometry=geo,
             size=size
         )
--- a/tests/test_sentinelhub_request.py
+++ b/tests/test_sentinelhub_request.py
@@ -446,6 +446,56 @@ class TestSentinelHubRequest(TestSentinelHub):
         with self.assertRaises(ValueError):
             SentinelHubRequest(**request_params)
 
+    def test_bbox_geometry(self):
+        """ Test intersection between bbox and geometry
+        """
+        evalscript = """
+            //VERSION=3
+
+            function setup() {
+                return {
+                    input: [{
+                        bands: ["B02", "B03", "B04"],
+                        units: "REFLECTANCE"
+                    }],
+                    output: {
+                        bands: 3
+                    }
+                };
+            }
+
+            function evaluatePixel(sample) {
+                return [2.5 * sample.B04, 2.5 * sample.B03, 2.5 * sample.B02];
+            }
+        """
+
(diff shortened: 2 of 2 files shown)


batic:
The error message when both `bbox` and `geometry` are not specified should change from `and/or` to just `or`:
```python
        if bbox is None and geometry is None:
            raise ValueError("'bbox' and/or 'geometry' have to be provided.")		            raise ValueError("'bbox' and/or 'geometry' have to be provided.")

```

batic:
Additionally, I would suggest checking if what we retrieve from SH is what we expect, checking  (corner) cases like what happens if we send to SH both bbox and geometry, but bbox only partially covers geometry (or not at all).

Moving these checks to unit-tests would then be appreciated.

AleksMat:
> The error message when both `bbox` and `geometry` are not specified should change from `and/or` to just `or`:
> 
> ```python
>         if bbox is None and geometry is None:
>             raise ValueError("'bbox' and/or 'geometry' have to be provided.")		            raise ValueError("'bbox' and/or 'geometry' have to be provided.")
> ```

Actually, you can still provide both and the service supports that. E.g. you provide a large bbox and a smaller geometry inside and the service should return an image of the entire bbox but with data only inside the geometry.

batic:
> > The error message when both `bbox` and `geometry` are not specified should change from `and/or` to just `or`:
> > ```python
> >         if bbox is None and geometry is None:
> >             raise ValueError("'bbox' and/or 'geometry' have to be provided.")		            raise ValueError("'bbox' and/or 'geometry' have to be provided.")
> > ```
> 
> Actually, you can still provide both and the service supports that. E.g. you provide a large bbox and a smaller geometry inside and the service should return an image of the entire bbox but with data only inside the geometry.

Ah, ok. In that case probably everything is covered by the changes @DomagojKorais has implemented.

DomagojKorais:
> Additionally, I would suggest checking if what we retrieve from SH is what we expect, checking (corner) cases like what happens if we send to SH both bbox and geometry, but bbox only partially covers geometry (or not at all).
> 
> Moving these checks to unit-tests would then be appreciated.

I will implement those additional checks.

DomagojKorais:
I prepared the unit tests, in order to cover all the combinations there are four of them:
* geometry contained into bbox
* bbox and geometry disjoint
* bbox and geometry intersects
* geometry contains bbox

One issue is that for each of them I need to make a new request, so I am not sure if this is optimal, since each request slows down the testing. 
Possible solutions are:
* avoid testing this aspect, since it is managed directly by SH.
* add those tests in other tests, in order to avoid making new requests.

AleksMat:
I think that the purpose of these tests should be that they make sure user parameters are correctly passed to the service and that is done by detecting statistical values in service responses. Therefore I suggest that we don't have tests whose single purpose is to verify that the service itself is working correctly. In the end, we don't want that these tests make to many requests to SH service as this only slows them down.

Cases of how SH service works given different relations between bbox and geometry IMO only test SH service itself. In this issue, I think that we only want to understand how they work but not really add them to tests.

So far almost all tests use only bbox and there is one test that uses geometry and a bbox, which is derived from geometry. My suggestion is that:

- `test_multipart_geometry` derives bbox from geometry. Let's remove bbox from this test and have only geometry.
- Let's add another test with bbox and geometry where these 2 are intersecting.

@batic what is your opinion about this? Would you want to have more tests?

batic:
> @batic what is your opinion about this? Would you want to have more tests?

Yes if we want to have "integration tests". Otherwise, no. 

I agree with your suggestion to have `test_multipart_geometry` just `geometry`, and adding one more example where both `bbox` and `geometry` are given but `bbox` does not completely cover the `geometry`.

DomagojKorais:
* `test_multipart_geometry` now uses only geometry and box is used only to compute the right size.
*  added test for the intersection case.

PR #164, opened 2021-02-19
BYOC API implementation.

batic:
Implementation and documentation for [BYOC api](https://docs.sentinel-hub.com/api/latest/reference/#tag/byoc_collection)


sentinelhub/sentinelhub_byoc.py
```diff
@@ -0,0 +1,264 @@
...
+    status: Optional[str] = None
+    tile_id: Optional[str] = field(metadata=dataclass_config(field_name='id'), default=None)
+    tile_geometry: Optional[Geometry] = field(metadata=geometry_config, default=None)
+    cover_geometry: Optional[Geometry] = field(metadata=geometry_config, default=None)
+    created: Optional[datetime] = field(metadata=datetime_config, default=None)
+    sensing_time: Optional[datetime] = field(metadata=datetime_config, default=None)
+    additional_data: Optional[dict] = None
+
+    @property
+    def tile_geometry_crs(self):
```
AleksMat, line 78: I'm just wondering how useful properties `tile_geometry_crs`, `cover_geometry_crs`, and `geometry` are. Usually you either have a geometry for all tiles or you don't. And if a geometry is empty you might still want to get a CRS or geometry.

Feel free to ignore this comment, I'm also fine if we leave them as they are.

AleksMat (commented):
Looks good, some comments about the notebook:

- There are a few typos in the notebook and I suggest spellchecking it. The way I usually do it is a bit hacky but it works:

  * building documention pages
  * open html file into which notebook has been parsed
  * copy-paste its content to somewhere where Grammarly could review it.

- You can put `;` at the end of line `tiles_gdf.plot(ax=ax)`

- In the evalscript having 

  ```
  function evaluatePixel(samples) {
    for (let i = 0; i < samples.length; i++) {
      var sample = samples[i];
      return [f*sample.B08, f*sample.B04, f*sample.B03, sample.dataMask];
    }
  ```
  
  should be completely the same as 
  
  ```
  function evaluatePixel(sample) {
    return [f*sample.B08, f*sample.B04, f*sample.B03, sample.dataMask];
    }
  ```
  
  In both cases you return the first sample.

sentinelhub/sentinelhub_byoc.py
batic: I wanted the `_crs` because they can be different between the tiles. `_geometry` seemed useful to convert to `shapely` object, but in the end one can always call `tile_geometry.geometry` or `cover_geometry.geometry` to achieve the same. As to some extent "less is better than more", I would remove them.

PR #165, opened 2021-02-22
Improved time utilities

AleksMat:
Changes:

- refactored `time_utils.py`
  * deprecated old functions
  * changed output type of `parse_time` and `parse_time_interval`
  * added `serialize_time`
- refactored unit tests for `time_utils.py`, now they are written with `pytest`
- updated code that deals with time
- fixed pylint `2.7.0` issues

Note that these improvements enable handling timestamps with microseconds.


sentinelhub/time_utils.py
```diff
@@ -147,34 +161,80 @@ def parse_time_interval(time):
     :param time: An input time
     :type time: str or datetime.datetime or (str, str) or (datetime.datetime, datetime.datetime)
     :return: interval of start and end date of the form `YYYY-MM-DDThh:mm:ss`
-    :rtype: (str, str)
+    :rtype: (datetime.datetime, datetime.datetime)
     :raises: ValueError
     """
     if time == SHConstants.LATEST:
```
batic, line 167: Why 1985?

sentinelhub/time_utils.py
```diff
@@ -147,34 +161,80 @@ def parse_time_interval(time):
...
 
-    if date_interval[1] < date_interval[0]:
+    if start_time > end_time:
         raise ValueError('Start of time interval is larger than end of time interval')
 
-    return date_interval
+    return start_time, end_time
+
+
+def serialize_time(timestamp_input, *, use_tz=False):
```
batic, line 190: I am not too keen on differences between parsing time and serializing time, namely, parsing works on single time while serializing can do a tuple. 

If that is the case, I'd prefer that `parse_time` (or perhaps `parse_datetime`) also works on a tuple. One can thus get rid of `parse_time_interval`. Perhaps it would be prudent to limit these functions to work on a single datetime or on a pair (to mimic time_interval). This would make the methods much more similar.

sentinelhub/aws.py
```diff
@@ -483,8 +484,8 @@ class will try to find the index automatically. If there will be multiple choice
         :type config: SHConfig or None
         """
         self.tile_name = self.parse_tile_name(tile_name)
-        self.datetime = self.parse_datetime(time)
-        self.date = self.datetime.split('T')[0]
+        self.timestamp = parse_time(time, ignoretz=True)
+        self.date_str = self.timestamp.isoformat().split('T')[0]
```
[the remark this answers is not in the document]
AleksMat: The problem is that the input parameter `time` can be either just a date or date + time. So then `self.timestamp` can be either `datetime.date` or `datetime.datetime`. On the other hand, `AwsProduct` can only obtain date information (without making additional requests). Because `AwsProduct` and `AwsTile` share some common methods then we have to define a date for both of them.

I agree this is a bit convoluted but for now, I would leave both a timestamp and a date object. I will however try storing date as a `datetime.date` object instead of a string.

sentinelhub/download/handlers.py
```diff
@@ -33,6 +33,7 @@ def new_download_func(self, request):
 def retry_temporal_errors(download_func):
     """ Decorator function for handling server and connection errors
     """
+    # pylint: disable=inconsistent-return-statements
```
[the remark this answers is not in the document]
AleksMat: I was thinking about that:

- Adding `return None` seemed to be a bit misleading since it will never come to this and we don't want to give an impression that this can return `None`.
- Moving `raise` statement outside the `for` loop as you suggested is problematic because `exception` variable is defined with try-except but now we would use it outside its scope. Of course, it would work but it is also not a good practice and pylint might raise a different kind of error because of it.

I think I now found the correct solution - the only way that we could come out of the for loop is when someone would set e.g. `self.config.max_download_attempts = 0`. Hence let's just raise the following:

```Python
raise DownloadFailedException('No download attempts available')
```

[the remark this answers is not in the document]
batic: 👍

sentinelhub/ogc.py
```diff
@@ -437,10 +438,9 @@ def get_dates(self):
             if not tile_info['properties']['date']:  # could be True for custom (BYOC) data collections
                 tile_dates.append(None)
             else:
-                tile_dates.append(datetime.datetime.strptime('{}T{}'.
-                                                             format(tile_info['properties']['date'],
-                                                                    tile_info['properties']['time'].split('.')[0]),
-                                                             '%Y-%m-%dT%H:%M:%S'))
+                date_str = tile_info['properties']['date']
+                time_str = tile_info['properties']['time']
+                tile_dates.append(parse_time(f'{date_str}T{time_str}'))
```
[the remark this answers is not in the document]
AleksMat: I think `parse_datetime` could be mistakenly understood that it parses only `datetime` objects. While `parse_time` perhaps signifies a bit more general use (although date always has to be present in some form).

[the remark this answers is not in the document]
batic: `parse_date_time` perhaps? Or is that already too convoluted?

sentinelhub/time_utils.py
AleksMat: Ah, this in order to get an entire interval of possibly available data. It is quite convoluted but it enables using `time='latest'` for OGC and Process API requests. Although now that I think about it OGC request will provide only the latest acquisition while Process API will provide a mosaic of acquisitions starting with the latest.

Perhaps it would be better to limit "latest" functionality only to OGC requests and remove this case from `parse_time_interval`?

sentinelhub/time_utils.py
AleksMat: I thought about this but the problem is that `parse_time_interval` is combining information about the start and end time while `serialize_time` is serializing each time independently. I.e.:

- `parse_time_interval(time_input)` is in most cases not the same as `parse_time(time_input[0]), parse_time(time_input[1])`
- `serialize_time((start_time, end_time))` is the same as `serialize_time(start_time), serialize_time(end_time)`.

Because of that, I don't think `parse_time` and `parse_time_interval` can be joined nicely.

sentinelhub/time_utils.py
batic: Limiting to OGC is a must, perhaps it would be prudent to remove it completely. Better that the users are explicit to what data they are looking for...

sentinelhub/ogc.py
```diff
@@ -177,7 +177,8 @@ def _get_wms_wcs_url_parameters(request, date):
                 seconds=0) else date - request.time_difference
             end_date = date if request.time_difference < datetime.timedelta(
                 seconds=0) else date + request.time_difference
-            params['TIME'] = '{}/{}'.format(start_date.isoformat(), end_date.isoformat())
+
+            params['TIME'] = f'{serialize_time(start_date, use_tz=True)}/{serialize_time(end_date, use_tz=True)}'
```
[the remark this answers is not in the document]
AleksMat: I think the cleanest (most explicit) solution is:

```Python
start_time, end_time = serialize_time((start_time, end_time), use_tz=True)
f'{start_time}/{end_time}'
```
I unified this for all 3 cases.

sentinelhub/ogc.py
```diff
@@ -234,13 +235,13 @@ def _get_fis_parameters(request, geometry):
         :return:  dictionary with parameters
         :rtype: dict
         """
-        date_interval = parse_time_interval(request.time)
+        start_time, end_time = serialize_time(parse_time_interval(request.time), use_tz=True)
```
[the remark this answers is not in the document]
AleksMat: You are right. I unified this.

sentinelhub/ogc.py
```diff
@@ -403,7 +404,7 @@ def _fetch_features(self):
             'BBOX': str(self.bbox.reverse()) if self.bbox.crs is CRS.WGS84 else str(self.bbox),
             'OUTPUTFORMAT': MimeType.get_string(MimeType.JSON),
             'SRSNAME': CRS.ogc_string(self.bbox.crs),
-            'TIME': '{}/{}'.format(self.time_interval[0], self.time_interval[1]),
+            'TIME': f'{self.time_interval[0]}/{self.time_interval[1]}',
```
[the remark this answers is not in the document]
AleksMat: You are right. I unified this.

sentinelhub/time_utils.py
batic: The fact that `parse_time_interval(time_input)` is not equal to `parse_time(time_input[0]), parse_time(time_input[1])` is the biggest issue I have. 

I know it allows nearly anything to be parsed, but that is a way to hiding explicitly providing `time_interval` from two dates/times (e.g., `start_time` and `end_time`). The case where `end_time == start_time`, is just a particular case. 

Validation of `time_interval` should be done elsewhere.

sentinelhub/aws.py
```diff
@@ -288,7 +289,7 @@ def has_reports(self):
         :return: `True` if the product has report xml files and `False` otherwise
         :rtype: bool
         """
-        return self.baseline > '02.05' or (self.baseline == '02.05' and self.date >= '2017-10-12')  # Not sure
+        return self.baseline > '02.05' or (self.baseline == '02.05' and self.date_str >= '2017-10-12')  # Not sure
```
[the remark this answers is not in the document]
AleksMat: Done. It turned out this change even uncovered a small bug in creating .SAFE structure for tile with baseline `02.05` and `date < '2017-10-12'`. :see_no_evil:

sentinelhub/aws.py
```diff
@@ -633,7 +620,7 @@ def get_tile_url(self, force_http=False):
         base_url = self.base_http_url if force_http else self.base_url
         url = '{}/tiles/{}/{}/{}/'.format(base_url, self.tile_name[0:2].lstrip('0'), self.tile_name[2],
                                           self.tile_name[3:5])
-        date_params = self.date.split('-')
+        date_params = self.date_str.split('-')
```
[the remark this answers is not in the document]
AleksMat: Updated.

sentinelhub/time_utils.py
```diff
@@ -63,8 +69,9 @@ def iso_to_datetime(date):
     :return: datetime instance
     :rtype: datetime
     """
+    warnings.warn('This function will be removed in the next version', category=SHDeprecationWarning)
```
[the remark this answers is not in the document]
AleksMat: I like the idea, but I think this might trigger quite a few changes which would get a bit out of the scope of this PR. So I'll open an internal issue about this and put it to "Wishlist".

sentinelhub/ogc.py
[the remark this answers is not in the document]
AleksMat: IMO, too verbose.

sentinelhub/time_utils.py
AleksMat: Ok, I'll look into this.

sentinelhub/ogc.py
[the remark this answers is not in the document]
batic: To some extent not importing functions, but modules, would improve this... Then one could call `time_utils.parse(...)` and the intent of the method would be completely clear ...

sentinelhub/time_utils.py
AleksMat: I see `parse_time_interval` as a transformation of various "loose" time definitions into a strict time interval definition with 2 `datetime.datetime` objects. Sometimes this is quite convenient.

For example, let's say you want an image for `2021-02-24`. With `parse_time_interval` you can make a Process API request with `time='2021-02-24'`. Without it you would have to write `time=('2021-02-24T00:00:00Z', '2021-02-24T23:59:59Z')`. For prototyping with sh-py, I think this soon becomes a bit annoying from the user's point of view.

Also, if we go further changing how `parse_time_interval` works I think this will break quite a few things in our codebase as well.

sentinelhub/download/handlers.py
```diff
@@ -58,6 +58,9 @@ def new_download_func(self, request):
                 time.sleep(sleep_time)
                 sleep_time *= backoff_coefficient
 
+        raise DownloadFailedException('No download attempts available - configuration parameter max_download_attempts '
```
batic, line 61: Any chance this part is reached because of some other (unhandled) exception? E.g. the `download_func` fails with "disk full" or sth? In such case the message is misleading.

AleksMat: Hm, good question but I'm quite sure that it is not possible. Our try-except is only catching `requests.RequestException` errors. Anything else would not be caught and therefore raise an error. The `except` segment also makes sure that it raises an error at the end of the last attempt. So the only way that the code can go past the `for` loop is if it never enters it.

sentinelhub/time_utils.py
AleksMat: After sleeping it over it is now clear to me that this is a good decision. So I removed support for `'latest'` from `parse_time_interval` and only added handling of this in `WebFeatureService` class. Now only WMS, WCS, and WFS functionalities support this option.

I wouldn't completely remove it since this has been a default option of `WmsRequest` and `WcsRequest` since the beginning and things would break if we changed that.

sentinelhub/ogc.py
```diff
@@ -351,15 +351,22 @@ def __init__(self, bbox, time_interval, *, data_collection=None, maxcc=1.0, data
...
         self.bbox = bbox
-        self.time_interval = parse_time_interval(time_interval)
+
+        self.latest_time_only = time_interval == SHConstants.LATEST
+        if self.latest_time_only:
+            self.time_interval = datetime.datetime(year=1985, month=1, day=1), datetime.datetime.now()
+        else:
+            self.time_interval = parse_time_interval(time_interval)
+
+        self.latest_time_only = time_interval == SHConstants.LATEST
```
batic, line 361: Duplicated line.

AleksMat: Ouch. :see_no_evil: 

Removed.

sentinelhub/time_utils.py
batic: I agree with the changes.

PR #166, opened 2021-02-23
FeatureIterator and more general improvements

AleksMat:
Changes:
- Implemented a general `FeatureIterator` class and unified feature iteration over Catalog, BYOC, Batch, WFS, and Geopedia services.
- Improvements in `SentinelHubDownloadClient` of how a thread lock and sessions are handled.
- Renamed "Processing API" to "Process API" everywhere. This way we'll be in line with SH service documentation.
- Fixed BYOC API reference links and added Catalog and Batch API reference links
- Removed a couple of outdated/unsupported Batch API parameters

Note that this PR builds on top of PR https://github.com/sentinel-hub/sentinelhub-py/pull/165. So that one needs to be reviewed and merged first.


batic:
I'll wait for the `time_utils` to get merged before reviewing this one.

AleksMat:
> I'll wait for the `time_utils` to get merged before reviewing this one.

Merge conflicts have been now resolved. This PR is ready for review.

sentinelhub/geopedia.py
```diff
@@ -388,15 +384,18 @@ def __len__(self):
     def _fetch_features(self):
         """ Retrieves a new page of features from Geopedia
         """
-        if self.next_page_url is None:
-            return
+        response = self.client.get_json(self.next, post_values=self.params, headers=self.gpd_session.session_headers)
```
batic, line 387: Are both `offset` and `limit` part of the `pagination['next']` link? Otherwise I think the `limit` gets lost...

sentinelhub/sentinelhub_byoc.py
```diff
@@ -93,17 +95,22 @@ def __init__(self, base_url=None, config=None):
...
+        `BYOC API reference <https://docs.sentinel-hub.com/api/latest/reference/#operation/getByocCollections>`_
 
         :return: iterator over collections
         """
         url = f'{self.byoc_url}/collections'
-        return iter_pages(client=self.client, service_url=url,
-                          exception_message='Failed to obtain information about available collections.')
+        return SentinelHubFeatureIterator(
+            client=self.client,
+            url=url,
```
batic, line 106: Perhaps `url` can be constructed inline, e.g.:
```python
   url = f'{self.byoc_url}/collections'
```

(similarly below). Personal preference for short lines. Feel free to ignore.

sentinelhub/sh_utils.py
```diff
@@ -1,40 +1,102 @@
...
+        return self
+
+    def __next__(self):
+        """ Method called to provide the next feature in iteration
+
+        :return: the next feature
+        """
+        while self.index >= len(self.features) and not self.finished:
+            new_features = self._fetch_features()
+            self.features.extend(new_features)
```
batic, line 52: I would suggest using `collections.deque` for the `features`. Appending entries to `collections.deque` (O(1)) should be better then extending a list (memory reallocation when allocated mem is exhausted), particularly for when there is a large number of features ...

sentinelhub/sh_utils.py
```diff
@@ -1,40 +1,102 @@
...
 from urllib.parse import urlencode
 
-from sentinelhub.config import SHConfig
-from sentinelhub.download import SentinelHubDownloadClient
 from sentinelhub.exceptions import MissingDataInRequestException
 
 
-def iter_pages(service_url, client=None, config=None, exception_message=None, **params):
-    """ Iterates over pages of items retrieved from Sentinel-Hub service
+class FeatureIterator(ABC):
```
batic, line 10: General comment about `FeatureIterators`... Is it possible to have any tests that are not "integration tests"?

sentinelhub/download/sentinelhub_client.py
```diff
@@ -61,9 +74,12 @@ def _execute_download(self, request):
                 LOGGER.debug('%s: Sleeping for %0.2f', thread_name, sleep_time)
                 time.sleep(sleep_time)
 
-    def _execute_with_lock(self, thread_unsafe_function, *args, **kwargs):
+    def _execute_thread_safe(self, thread_unsafe_function, *args, **kwargs):
         """ Executes a function inside a thread lock and handles potential errors
         """
+        if self.lock is None:
```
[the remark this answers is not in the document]
AleksMat: No, for a lock to be effective you have to create `1` lock and use it for all threads. If you create and use a different lock for each thread then such locks won't have any effect.

The idea is that a lock is created in `SentinelHubDownloadClient.download` method, which is the only one that has to download multiple things at once. Then here each thread can use the same `self.lock` object.

I see how this could be misinterpreted but if here `self.lock` is `None` this means that we didn't have to create any locks and therefore aren't using multiple threads. Hence we can just call the function directly.

sentinelhub/geopedia.py
```diff
@@ -27,27 +28,27 @@ def __init__(self, config=None):
...
-            return layer
+def _parse_geopedia_layer(layer, return_wms_name=False):
+    """ Helper function for parsing Geopedia layer name. If WMS name is required and wrong form is given it will
+    return a string with 'ttl' at the beginning. (WMS name can also start with something else, e.g. only 't'
+    instead 'ttl', therefore anything else is also allowed.) Otherwise it will parse it into a number.
+    """
+    if not isinstance(layer, (int, str)):
+        raise ValueError("Parameter 'layer' should be an integer or a string, but {} found".format(type(layer)))
+
+    if return_wms_name:
```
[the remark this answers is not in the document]
AleksMat: But it is not just that. This function can either parse layer name into a layer id, e.g. `12345`, or it can parse it into a string identifier for WMS requests, e.g. `'ttl12345'`. And naming the parameter `return_as_string` does not reflect that.

Note that in this PR I just changed this from a method to a function. I didn't change its content.

sentinelhub/geopedia.py
AleksMat: They do, I checked. But I agree it is a bit scary if this would ever accidentally get changed on Geopedia.

sentinelhub/config.py
```diff
@@ -263,7 +263,7 @@ def get_sh_oauth_url(self):
         return '{}/oauth/token'.format(self.sh_base_url)
 
     def get_sh_processing_api_url(self):
-        """  Provides URL for Sentinel Hub processing API endpoint
+        """  Provides URL for Sentinel Hub Process API endpoint
```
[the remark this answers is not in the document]
AleksMat: Renamed

sentinelhub/download/client.py
```diff
@@ -168,7 +168,7 @@ def get_json(self, url, post_values=None, headers=None, request_type=None, **kwa
         request = DownloadRequest(url=url, headers=json_headers, request_type=request_type, post_values=post_values,
                                   data_type=MimeType.JSON, **kwargs)
 
-        return self.download(request)
+        return self._single_download(request, True)
```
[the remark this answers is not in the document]
AleksMat: Done

sentinelhub/download/client.py
```diff
@@ -180,7 +180,7 @@ def get_xml(self, url, **kwargs):
         :rtype: XML instance or None
         """
         request = DownloadRequest(url=url, data_type=MimeType.XML, **kwargs)
-        return self.download(request)
+        return self._single_download(request, True)
```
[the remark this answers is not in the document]
AleksMat: Done

sentinelhub/config.py
```diff
@@ -286,3 +286,13 @@ def get_sh_rate_limit_url(self):
         :rtype: str
         """
         return '{}/aux/ratelimit'.format(self.sh_base_url)
+
+    def raise_missing_instance_id(self):
```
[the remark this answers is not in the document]
AleksMat: Done. The name is now even longer but it is more undestandable.

sentinelhub/download/sentinelhub_client.py
```diff
@@ -87,17 +103,29 @@ def _prepare_headers(self, request):
...
+    def _get_session_headers(self):
+        """ Provides up-to-date session headers
+
+        Note that calling session_headers property triggers update if session has expired therefore this has to be
+        called in a thread-safe way
+        """
+        return self.get_session().session_headers
+
+    def get_session(self):
+        """ Provides the session object used by the client
```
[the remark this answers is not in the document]
AleksMat: Yes, technically we can do that. Done

sentinelhub/sentinelhub_byoc.py
AleksMat: Yes, since it is a trivial operation we can have it directly in the constructor of `SentinelHubFeatureIterator`. Done

sentinelhub/sh_utils.py
AleksMat: The time complexity for extending a list and extending a `deque` is the same. You can check that [here](https://wiki.python.org/moin/TimeComplexity). They both have dynamic memory allocation, probably similar to Vectors in Java and C++.

sentinelhub/sh_utils.py
AleksMat: Sure, I can implement a dummy `_fetch_features` method and prepare a few tests. Will do in the morning.

sentinelhub/config.py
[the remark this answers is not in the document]
batic: To be shorter, as `raise_for_status` is not `raise_for_wrong/missing/bad/_status`, this method could simply be `raise_for_instance_id`.

sentinelhub/geopedia.py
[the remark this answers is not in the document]
batic: Hard to swallow, but I don't have better solution (apart perhaps from having two functions). Leave it as is.

sentinelhub/sh_utils.py
batic: Yes, amortised extensions have the same complexity. Expanding a list, when list would outgrown its current memory allocation, consists of copying the whole array to a new (contiguous) block of memory. If original list is already large, this might take quite some time. `dequeue.extend` does not work with contiguous memory allocations, which means that there is no need for any memory reallocations. 

On the other hand, we probably won't be downloading the whole internet with these, so until we notice a performance hit, let's keep features as `list`.

sentinelhub/geopedia.py
batic: Perhaps add a check and raise an error?

sentinelhub/config.py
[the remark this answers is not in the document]
AleksMat: Hm, let's just leave the longer name as it specifically checks for missing credentials and not for wrong/bad credentials.

sentinelhub/geopedia.py
AleksMat: I added a check that verifies these parameters exist in the next page URL.

sentinelhub/sh_utils.py
AleksMat: Added some unit tests that verify the basic mechanics of `FeatureIterator` class is working correctly.

batic (approved):
Ready to be merged.

PR #167, opened 2021-02-24
configuration documentation updates

batic:


AleksMat (commented):
I checked the changes and I like all of them. I would also recommend the following:

- At the beginning of the `Sentinel Hub Configuration` chapter you can add a paragraph stating that one has to register to SH in order to obtain credentials which can be done with a trial account, payable account, or a free non-commercial account sponsored by ESA. Then you can include information written in the first paragraph [here](https://sentinelhub-py.readthedocs.io/en/latest/examples/ogc_request.html#Sentinel-Hub-account). Although the link to ESA page would have to be updated with a new OSEO call.
- Go through all example notebooks and replace paragraphs about credentials with a sentaince that tells users that they can find all information about this here in configure.html.

AleksMat (commented):
Looks good, here are some comments and suggestions:

- In OGC notebook there is still `1` link to `processing_api_request.ipynb` which has to be updated. Next to it there is also a string `Processing API examples` which should also be updated. Don't know how we both missed that.
- In multiple places in notebooks we have a link to documentation `http://sentinelhub-py.readthedocs.io/...`. I suggest changing `http` to `https` since readthedocs works on `https` now.
- I agree with reverting back `jpeg` to `jpg`. Otherwise, a fix for that AWS preview would be a bit hacky. At the end of the day, we never use jpg ourselves, and for those users who do let's keep it the way it was.
- As mentioned, we could add redirects from the old naming of Process API documentation notebook page to the new one. It seems this can be quickly handled either on a level of ReadTheDocs or with a Sphinx extension. However, we would have test if it really works. Perhaps I can prepare a minor PR on top of this one to add this redirect?

batic:
I agree with everything. Will switch http -> https and update OGC notebook. Let's merge afterwards and have a look at the redirect once it is merged.

docs/source/conf.py
```diff
@@ -223,3 +224,8 @@
     shutil.copytree('../../examples', './examples')
 except FileExistsError:
     pass
+
+# redirects
+redirects = {
```
AleksMat, line 229: I found a solution for this. It turns out redirects don't work if a file, from which to redirect, doesn't exist. So let's create an empty file from which to redirect from:

```Python
# redirects
redirects = {
    'examples/processing_api_request': 'process_request.html',
}
# create empty files to redirect from
for file_path in redirects:
    file_path = f'{file_path}.rst'
    open(file_path, 'w').close()
```
I can confirm this works. Perhaps in the future, they will improve this...

AleksMat (commented):
Everything looks fine and I even found a fix for redirect. :smiley:

docs/source/conf.py
batic: 👍

PR #169, opened 2021-03-03
Open time intervals and "cloudy" collections

batic:
This PR tweaks the time (interval) parsing in order to allow open-ended interval requests to (some) SH services. Additionally, it adds `has_cloud_coverage(=False)` flag to DataCollections and sets it to true to collections that have CC as queryable parameter.

### DIFF
--- a/sentinelhub/data_collections.py
+++ b/sentinelhub/data_collections.py
@@ -152,6 +152,7 @@ class DataCollectionDefinition:
     bands: Tuple[str, ...] = None
     collection_id: str = None
     is_timeless: bool = False
+    has_cloud_coverage: bool = False
 
     def __post_init__(self):
         """ In case a list of bands has been given this makes sure to cast it into a tuple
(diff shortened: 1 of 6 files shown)


sentinelhub/data_collections.py
```diff
@@ -152,6 +152,7 @@ class DataCollectionDefinition:
     bands: Tuple[str, ...] = None
     collection_id: str = None
     is_timeless: bool = False
+    has_cloud_coverage: bool = False
```
AleksMat, line 155: Hm, pylint is complaining about too many attributes -> [here](https://travis-ci.org/github/sentinel-hub/sentinelhub-py/jobs/761298763#L360).

I don't see a better solution but to suppress pylint for this class: `# pylint: disable=too-many-instance-attributes`

AleksMat (commented):
Look good, there is only 1 small pylint issue.

sentinelhub/data_collections.py
batic: Added.

PR #172, opened 2021-03-18
Small changes

batic:
This PR adds `DataCollection` related query parameters to `SentinelHubCatalog.search`, removes `datetime` parameter if not provided (to search through timeless data), and allows user to override mime type of SentinelHubRequest. 

All these changes will be needed for upcoming changes in eo-learn.


AleksMat (approved):
Looks good.

PR #175, opened 2021-03-24
demInstance parameter for DEM collections.

batic:
This PR implements dem_instance parameter for DEM collections.

### DIFF
--- a/sentinelhub/data_collections.py
+++ b/sentinelhub/data_collections.py
@@ -21,7 +21,7 @@ class _CollectionType:
     SENTINEL1 = 'Sentinel-1'
     LANDSAT8 = 'Landsat 8'
     MODIS = 'MODIS'
-    DEM = 'Mapzen DEM'
+    DEM = 'DEM'
     BYOC = 'BYOC'
     BATCH = 'BATCH'
     LANDSAT5 = 'Landsat 5'
@@ -258,12 +258,20 @@ class DataCollection(Enum, metaclass=_DataCollectionMeta):
 
     DEM = DataCollectionDefinition(
         api_id='DEM',
-        wfs_id='DSS4',
         collection_type=_CollectionType.DEM,
         bands=_Bands.DEM,
-        is_timeless=True,
+        is_timeless=True
+    )
+    DEM_MAPZEN = DEM.derive(
         dem_instance='MAPZEN'
     )
+    DEM_COPERNICUS_30 = DEM.derive(
+        dem_instance='COPERNICUS_30'
+    )
+    DEM_COPERNICUS_90 = DEM.derive(
+        dem_instance='COPERNICUS_90'
+    )
+
     MODIS = DataCollectionDefinition(
         api_id='MODIS',
         catalog_id='modis',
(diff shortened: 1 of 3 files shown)


sentinelhub/data_collections.py
```diff
@@ -372,6 +375,8 @@ def define(cls, name, *, api_id=None, catalog_id=None, wfs_id=None, service_url=
         :type collection_id: str or None
         :param is_timeless: `True` if a data collection can be filtered by time dimension and `False` otherwise
         :type is_timeless: bool
+        :param dem_instance: DEM instance for DEM collections (one of MAPZEN, COPERNICUS_30, COPERNICUS_90)
```
AleksMat, line 378: Instead of listing the options, you can perhaps reference the documentation:
```
one of the options listed in `DEM documentation <https://docs.sentinel-hub.com/api/latest/data/dem/#deminstance>`__
```

sentinelhub/data_collections.py
```diff
@@ -372,6 +375,8 @@ def define(cls, name, *, api_id=None, catalog_id=None, wfs_id=None, service_url=
...
+        :type dem_instance: str
```
AleksMat, line 379: `str or None`

sentinelhub/sentinelhub_catalog.py
```diff
@@ -188,6 +188,9 @@ def _get_data_collection_filters(data_collection):
         if data_collection.timeliness:
             filters['timeliness'] = {'eq': data_collection.timeliness}
 
+        if data_collection.dem_instance:
+            filters['demInstance'] = data_collection.dem_instance
```
AleksMat, line 192: If I'm not mistaken Catalog API doesn't support DEM. If it ever will then the query would probably be something like: `{'eq': data_collection.dem_instance}`.

I suggest that for now this is removed. If support gets added we can add the correct filter then.

AleksMat (commented):
- I'm wondering if it makes sense to add predefined enums `DataCollection.DEM_COPERNICUS_30` and `DataCollection.DEM_COPERNICUS_90` so that users don't have to do that?

- A bit off-topic but since we are already working on `DataCollection` perhaps we can also improve the error message we discussed earlier. This can be done simply by replacing [this line](https://github.com/sentinel-hub/sentinelhub-py/blob/feat/dem-instance/sentinelhub/data_collections.py#L438) with the following:

```Python
existing_collection = cls._value2member_map_[definition]
raise ValueError(f'This data collection definition is already taken by {existing_collection}. Two different '
                 f'DataCollection enums cannot have the same definition')
```

batic:
> I'm wondering if it makes sense to add predefined enums DataCollection.DEM_COPERNICUS_30 and DataCollection.DEM_COPERNICUS_90 so that users don't have to do that?

Was wondering that myself. In eo-learn DemTask I just construct a collection based on dem_instance (from existing DataCollection.DEM), but perhaps it would be better to have them specified. 

I vote yes. You?

AleksMat:
> I vote yes. You?

Yes, I'm ok with adding them.

Issue #116, opened 2020-03-05
Error downloading L2A .SAFE products from start of 2017


jirikadlec2:
When I try to use sentinelhub.aws to download L2A .SAFE products from January-April 2017, sentinelhub.aws fails with an error.

Example command:
`sentinelhub.aws --product S2A_MSIL2A_20170414T101021_N0001_R022_T33UUQ_20190508T121145 --l2a`

Error message:
[...]

PR #181, opened 2021-06-14
Fix for issue #116

marcosdejesus:
This PR address issue #116

 This bug fix insert some checks to build the SAFE structure of reprocessed Sentinel-2 L2A products appropriately.

### DIFF
--- a/sentinelhub/aws_safe.py
+++ b/sentinelhub/aws_safe.py
@@ -237,7 +237,7 @@ class SafeTile(AwsTile):
         if self.data_collection is DataCollection.SENTINEL2_L2A:
             for mask in AwsConstants.CLASS_MASKS:
                 for resolution in [AwsConstants.R20m, AwsConstants.R60m]:
-                    if self.baseline <= '02.06':
+                    if self.baseline <= '02.06' or self.baseline == '00.01':
                         mask_name = self.get_img_name(mask, resolution)
                     else:
                         mask_name = self.get_qi_name('{}PRB'.format(mask), resolution.lstrip('R'), MimeType.JP2)
--- a/sentinelhub/aws_safe.py
+++ b/sentinelhub/aws_safe.py
@@ -265,7 +265,7 @@ class SafeTile(AwsTile):
         tree = client.get_xml(self.get_url(AwsConstants.METADATA))
 
         tile_id_tag = 'TILE_ID_2A' if (self.data_collection is DataCollection.SENTINEL2_L2A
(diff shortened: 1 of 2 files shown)


sentinelhub/aws_safe.py
```diff
@@ -237,7 +237,7 @@ def get_safe_struct(self):
         if self.data_collection is DataCollection.SENTINEL2_L2A:
             for mask in AwsConstants.CLASS_MASKS:
                 for resolution in [AwsConstants.R20m, AwsConstants.R60m]:
-                    if self.baseline <= '02.06':
+                    if self.baseline <= '02.06' or self.baseline == '00.01':
```
AleksMat, line 240: It seems that everywhere else baseline `00.01` is handled the same as baselines `02.07` and above. Only here baseline `00.01` is handled the same as baselines `02.06` and below. Is there a reason for this?

Note that this results into a different naming of 4 metadata masks.

- this is what we get at the moment:
```
T33UUQ_20170414T101021_SNW_20m.jp2
T33UUQ_20170414T101021_SNW_60m.jp2
T33UUQ_20170414T101021_CLD_20m.jp2
T33UUQ_20170414T101021_CLD_60m.jp2
```
- this is what we could get for baselines `>=02.07`
```
MSK_SNWPRB_20m.jp2
MSK_SNWPRB_60m.jp2
MSK_CLDPRB_20m.jp2
MSK_CLDPRB_60m.jp2
```
- and this is what we could get for baselines `<=02.06` (if everything would be consistent with those):
```
L2A_T33UUQ_20170414T101021_SNW_20m.jp2
L2A_T33UUQ_20170414T101021_SNW_60m.jp2
L2A_T33UUQ_20170414T101021_CLD_20m.jp2
L2A_T33UUQ_20170414T101021_CLD_60m.jp2
```

So the current naming of these masks for `00.01` is not consistent neither with the old baselines nor with the newer.

AleksMat (commented):
Looks great. I only have one comment/question that I would like to discuss before we merge.

sentinelhub/aws_safe.py
marcosdejesus: You are completely right.

I just changed this part to treat the reprocessed files as baselines `02.07` and above.

AleksMat: Excellent, all clear now.

AleksMat:
Thanks for the contribution. It will be included in today's release. :+1:

PR #183, opened 2021-06-15
New Landsat collections

AleksMat:
Changes:
- added `DataCollection` enums for new Landsat collections,
- updated tests to use new collections,
  * so far I haven't added any new OGC/Process API tests - more could be added in the next iterations if needed
- updated some example notebooks
  * I intentionally didn't update FIS and OGC example notebooks as it would be good that first Python Scripts configuration template gets updated.

I hope the naming of the new collections is ok. I went with `LANDSAT45_L1` as using `LANDSAT4-5_L1` or `LANDSAT4_5_L1` seemed weird.


zigaLuksic (approved):
Seem allright

PR #185, opened 2021-07-01
DataCollection transfer fix

AleksMat:
This solves the problem with the following procedure:

- define a custom `DataCollection` enum,
- pass the new enum to another process (by `pickle` serialization),
- deserialize the enum in the new process.

This enables that this object is used with Ray framework.


zigaLuksic (approved):
Seems ok for now.

PR #186, opened 2021-07-05
added byoc notebook imporvements

ChungXiangHong:
Some improvements have been added to the `byoc_request.ipynb`.
Main differences:
- The sequence of steps in `BYOC collections` and `BYOC tiles` has been adjusted. We first create a new collection, then rename the collection to show how to update the collection, and finally we delete this empty collection.
- Using `get_collection` and `get_tile` to obtain the collection and the tile we created in the notebook, which prevent users from accidentally editing their existing collections and tiles. `next` and `list` are still described in the md cell as other alternatives to interact with collections and tiles.
- In order to actually demonstrate how to ingest tiles from AWS S3 bucket, we take Sentinel-2 L2A 120m Mosaic open registry data on AWS as an example, which allow user following this notebook to produce the exact same output of the false color image of Caspian Sea.


AleksMat (changes requested):
I checked your PR. It adds some nice improvements to the notebook and in general I agree with them. There are just various small style issues. Unfortunately GitHub won't allow me to write comments next to the code of this PR. This is because images in the notebook have changed and that causes that the overall change is too for GitHub to display the code. So I'm just writing all my comments here:

- `import os` is redundant

- I don't particularly like how the 2nd cell, where users have to insert credentials, is written. (I also didn't like the previous version.) It is because it doesn't take into account that users could have their credentials saved in config.json or, in case of AWS credentials, even in `~/.aws/credentials` file. Instead it forces them to write credentials into the notebook. I recommend something like:

```python
# Insert your credentials here in case you don't already have them in config.json file:
SH_CLIENT_ID = ''
SH_CLIENT_SECRET = ''
AWS_ACCESS_KEY_ID = ''
AWS_SECRET_ACCESS_KEY = ''


if SH_CLIENT_ID and SH_CLIENT_SECRET:
    config.sh_client_id = SH_CLIENT_ID
    config.sh_client_secret = SH_CLIENT_SECRET
    
if not config.sh_client_id or not config.sh_client_secret:
    print("Warning! To use Sentinel Hub BYOC API, please provide the credentials (client ID and client secret).")
    
if AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY:
    config.aws_access_key_id = AWS_ACCESS_KEY_ID
    config.aws_secret_access_key = AWS_SECRET_ACCESS_KEY
```

I wouldn't even raise a warning if AWS credentials are not set because they can have them in `~/.aws/credentials` file.

- No need to define too many new variables if they will only be used in the next line. Instead of:
```python
created_collection_name = created_collection['name']
created_collection_id = created_collection['id']
print('name:', created_collection_name)
print('id:', created_collection_id)
```
you can just do
```python
print('name:', created_collection['name'])
print('id:', created_collection['id'])
```

- Instead of
```python
my_collection_using_next = next(collections_iterator)
my_collection_using_next
print('name:', my_collection_using_next['name'])
print('id:', my_collection_using_next['id'])
```
you can remove the redundant line:
```python
my_collection_using_next = next(collections_iterator)

print('name:', my_collection_using_next['name'])
print('id:', my_collection_using_next['id'])
```

- This note

"Note: next() will only show the first collection on its first execution, the second collection for its second execution, and so on. if you already have collections the output above won’t necessarily show the one we just created." 

mostly just describes how an iterator and `next` function work. Perhaps the part worth mentioning is only that `collections_iterator` won't necessarily return collections in the same order they were created.

- You have: "Below we’re going to rename the ‘byoc-tutorial’ collection to ‘byoc-tutorial-renamed’:" but then you rename `'new collection'` into `'renamed new collection'`.

- Typo in "Trying to access this collection now will fali." -> "fail"

- In sentence:

"users can have a look at the documentation or use the BYOC tool that will take care of creating a collection and ingesting the tiles for you."

there used to be links to documentation but there are not anymore there. Any reason for removing them?

- Instead of "Sentinel-Hub" it should be "Sentinel Hub" everywhere. I see that already the previous version had this mistake.

- Where you have a note with bullet points:

```
**Note:**
- The most important field of the tile is its `path` on an s3 bucket. ...
```
make sure to add an empty line:

```
**Note:**

- The most important field of the tile is its `path` on an s3 bucket. ...
```
otherwise it won't parse into documentation correctly.

- In function signature:

```python
list_objects_path(bucket, aws_access_key_id, aws_secret_access_key, y=None, m=None, d=None):
```
I recommend renaming `y`, `m`, `d` into something more descriptive, e.g. `year_count`, `month_count`, `day_count`. Single-letter names can here be even a bit misleading as one would think you have to specify current year, month and day. It also doesn't make sense to have default values `None` for them as such values wouldn't work. I also recommend simply passing `config` object instead of both parameters for AWS credentials. So the signature could be:

```python
def list_objects_path(bucket, year_count, month_count, day_count, config):
```

- Line:
```python
tiles_path = list_objects_path('sentinel-s2-l2a-mosaic-120', config.aws_access_key_id, config.aws_secret_access_key, y=1, m=1, d=1)
```
is quite long, so I recommend splitting it into
```python
tiles_path = list_objects_path(
    'sentinel-s2-l2a-mosaic-120',
    config.aws_access_key_id,
    config.aws_secret_access_key,
    y=1,
    m=1,
    d=1
)
```

- Indentation in:
```python
    byoc_tile = ByocTile(path=f'{tile}(BAND).tif', 
                         sensing_time=dt.datetime(int(tile.split('/')[0]), 
                                                  int(tile.split('/')[1]), 
                                                  int(tile.split('/')[2]), 
                                                  0, 
                                                  0, 
                                                  0)
                        )
```
is a bit too much. I suggest changing to something like:

```python
year, month, day = tile.split('/')[:3]
byoc_tile = ByocTile(
    path=f'{tile}(BAND).tif', 
    sensing_time=dt.datetime(int(year), int(month), int(day))
)
```
- You can simplify sentence "To request one specific tile in the collection, you can do it by passing the collection id and the tile id to the `get_tile` method." into "To request a specific tile you can pass a collection id and a tile id to the `get_tile` method."

- There are `4` cases in the code where you write something like:

```python
[x['path'] for x in tiles if x['status']=='FAILED']
```
I recommend choosing a better name for variable `x` and also mind the spaces around `==` operator:
```python
[tile['path'] for tile in tiles if tile['status'] == 'FAILED']
```

- This was here from before, but line:

```python
tiles_gdf = gpd.GeoDataFrame(tiles_for_visualized, geometry=[t.cover_geometry.transform(CRS.WGS84).geometry for t in tiles_for_visualized], crs='epsg:4326')
```
is also a bit too long so I recommend splitting it:

```python
tiles_gdf = gpd.GeoDataFrame(
    tiles_for_visualized,
    geometry=[tile.cover_geometry.transform(CRS.WGS84).geometry for tile in tiles_for_visualized],
    crs='epsg:4326'
)
```

- In the sentence " ... the ingested tiles are 100 tiles from the pull out from the Sentinel-2 L2A 120m Mosaic ..." I think you can remove "from the pull out " part as it is a bit confusing.

- In the following there is a typo and a missing empty line which causes bad parsing into documentation:

```
Updating and deleting a tile follow the same logic as updating/deleting a collection:
- To updatea tile:
```

I recommend:

```
Updating and deleting a tile follow the same logic as updating/deleting a collection.

- To update a tile:
```

- Typo in "Alternatively using dataclasee:"

- Another mistake which we already had in the previous version: The following has too much indentation:

```python
request = SentinelHubRequest(
        evalscript=false_color_evalscript,
        input_data=[
            SentinelHubRequest.input_data(
                data_collection=data_collection,
                time_interval=tile_time
            )
        ],
        responses=[
            SentinelHubRequest.output_response('default', MimeType.PNG)
        ],
        bbox=caspian_sea_bbox,
        size=bbox_to_dimensions(caspian_sea_bbox, 100),
        config=config
    )
```
it should be:

```python
request = SentinelHubRequest(
    evalscript=false_color_evalscript,
    input_data=[
        SentinelHubRequest.input_data(
            data_collection=data_collection,
            time_interval=tile_time
        )
    ],
    responses=[
        SentinelHubRequest.output_response('default', MimeType.PNG)
    ],
    bbox=caspian_sea_bbox,
    size=bbox_to_dimensions(caspian_sea_bbox, 100),
    config=config
)
```

ChungXiangHong:
@AleksMat I've addressed the styling issue and corrected the typos. Links used to be in the sentence "users can have a look at the documentation or use the BYOC tool that will take care of creating a collection and ingesting the tiles for you" are added back (I accidentally forgot to put the links).

Thank you for your time and comments, I learned a lot it.

AleksMat (approved):
Excellent, I can confirm all issues have been resolved. :+1:

PR #191, opened 2021-07-20
An option to hide config credentials

AleksMat:
Instances of `SHConfig` are usually passed around in the code a lot, which means that some frameworks, such as eo-learn, would log it and write into a report. But the problem is that a representation of a config object reveals all its parameters, including its credentials.

This PR adds an option to hide credentials. It also refactors unit tests for `SHConfig` and adds a few new ones.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,71 +1,109 @@
-import unittest
+"""
+Unit tests for config.py module
+"""
 import json
 import os
 
-from sentinelhub import SHConfig, TestSentinelHub
+import pytest
+
+from sentinelhub import SHConfig
 
 
-class TestSHConfig(TestSentinelHub):
-    def test_configuration(self):
-        SHConfig().save()
+def test_config_file():
+    config = SHConfig()
 
-        config_file = SHConfig().get_config_location()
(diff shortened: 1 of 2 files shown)


tests/test_config.py
```diff
@@ -1,71 +1,109 @@
...
+    with pytest.raises(ValueError):
         config.save()
 
-    def test_raise_for_missing_instance_id(self):
-        config = SHConfig()
+    new_value = 150.5
+    config.download_timeout_seconds = new_value
+    config.save()
+    config = SHConfig()
+    assert config.download_timeout_seconds == new_value, 'Saved value should have changed'
```
zigaLuksic, line 61: Since this can fail, the *restore* in the following lines could not happen. One way to avoid this could be through using fixtures in the same way as when  testing ray clusters in eo-learn (so that the fixture restores the SHConfig after the test).

AleksMat: I agree and added the fixture. I made sure the config is restored even if a test failed.

zigaLuksic: Perfect!

zigaLuksic (approved):

PR #203, opened 2021-09-08
Make Mimetype `from_string` parse `get_string` output

zigaLuksic:
Makes the two operations pseudo-inverses of each other.

### DIFF
--- a/sentinelhub/constants.py
+++ b/sentinelhub/constants.py
@@ -342,15 +342,22 @@ class MimeType(Enum):
         :return: A mime type enum
         :rtype: MimeType
         """
+        if '/' in mime_type_str:
+            # handles strings of shape 'image/jpeg'
+            mime_type_str = mime_type_str.split('/')[-1]
+
         if MimeType.has_value(mime_type_str):
             return MimeType(mime_type_str)
 
         try:
             return {
                 'tif': MimeType.TIFF,
+                'plain': MimeType.TXT,
                 'jpeg': MimeType.JPG,
                 'jpg': MimeType.JPG,
+                'x-tar': MimeType.TAR,
                 'hdf5': MimeType.HDF,
+                'x-hdf': MimeType.HDF,
                 'h5': MimeType.HDF
             }[mime_type_str]
         except KeyError as exception:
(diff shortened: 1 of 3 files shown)


sentinelhub/constants.py
```diff
@@ -342,15 +342,22 @@ def from_string(mime_type_str):
         :return: A mime type enum
         :rtype: MimeType
         """
+        if '/' in mime_type_str:
```
AleksMat, line 345: Before you do this, I think you can try calling `mimetypes.guess_extension`. This will handle all cases such as `text/plain`, `application/x-tar`, `application/x-hdf`, etc. Otherwise it seems weird that this is being handled in the dictionary below. If an input of this method is `plain` you probably don't want to parse it into `MimeType.TXT`, only if it is `text/plain`.

In case `mimetypes.guess_extension` returns `None`, then you try splitting by `'/'`. A case for that would be `application/tar`, which is for some applications a valid mime type but `mimetypes` doesn't acknowledge that.

sentinelhub/constants.py
```diff
@@ -342,28 +342,22 @@ def from_string(mime_type_str):
...
         try:
             return {
                 'tif': MimeType.TIFF,
                 'jpeg': MimeType.JPG,
-                'jpg': MimeType.JPG,
                 'hdf5': MimeType.HDF,
                 'h5': MimeType.HDF
             }[mime_type_str]
         except KeyError as exception:
             raise ValueError(f'Data format .{mime_type_str} is not supported') from exception
```
AleksMat, line 359: `.{mime_type_str}` should be now changed into `{mime_type_str}` in the error message. Otherwise the message could be misleading.

sentinelhub/constants.py
```diff
@@ -342,28 +342,22 @@ def from_string(mime_type_str):
         :return: A mime type enum
         :rtype: MimeType
         """
+        guessed_extension = mimetypes.guess_extension(mime_type_str)
+        mime_type_str = guessed_extension.strip('.') if guessed_extension else mime_type_str
+
```
AleksMat, line 347: I would prefer if you left in that splitting by `'/'`. That would handle case `application/tar`.

I think this splitting should be important also to simplify [this line](https://github.com/sentinel-hub/sentinelhub-py/blob/master/sentinelhub/sentinelhub_request.py#L48)

PR #204, opened 2021-09-09
Add option to enable progress bar when downloading

zigaLuksic:


sentinelhub/download/client.py
```diff
@@ -1,13 +1,14 @@
...
 """
-import concurrent.futures
 import logging
 import warnings
 import os
 import sys
+from concurrent.futures import ThreadPoolExecutor, as_completed
 
 import requests
+from tqdm import tqdm
```
AleksMat, line 11: In the past it used to be important to do `from tqdm.auto import tqdm` in order to ensure that the progress bar also works in Jupyter notebooks. I don't know if this is still the case.

AleksMat (commented):
I suggest including `progress_bar=True` somewhere in tests. The perfect candidate is [this test](https://github.com/sentinel-hub/sentinelhub-py/blob/master/tests/test_download.py#L97).  This test is also supposed to verify the order of downloaded responses, although it might not be very accurate with only 3 items.

zigaLuksic:
> I suggest including `progress_bar=True` somewhere in tests. The perfect candidate is [this test](https://github.com/sentinel-hub/sentinelhub-py/blob/master/tests/test_download.py#L97). This test is also supposed to verify the order of downloaded responses, although it might not be very accurate with only 3 items.

I can try adding an additional test that verifies the order? Would probably requires some mocking though :thinking:
I enabled it in this test to try it out, turns out one has to run it with `pytest -s` for the progress bars to display. How would you check whether the progress bar is displayed? Or do we just check that things dont fail if this is turned on?

sentinelhub/download/client.py
zigaLuksic: I can just change this since it's a minor change.

AleksMat: Yes, I can now confirm there is still this distinction. So we should use `tqdm.auto`.

AleksMat:
For now we can also just add it so that it checks things don't fail. I'm not sure how to correctly check that `tqdm` is properly displayed.

I remembered that our eo-learn `SentinelHubInputTask` tests somewhat verify the order and those still pass. Should be fine for now.

PR #205, opened 2021-09-09
Multiple minor improvements

AleksMat:
This PR does the following:

- removes deprecated functions and paramters that have been deprecated for a year or more,
- fixes an issue where Process API, Statistical API and Catalog API were incorrectly sending CRS info together with geometry - it adds option to convert to geojson without CRS,
- adds support for `identifier` parameter for Process and Statistical APIs (i.e. parameter used for data fusion)
  * I intentionally used name `identifier` instead of `id`, which is used by the service, as `id` is a reserved word in Python.


zigaLuksic (approved):
Seems great :)

PR #207, opened 2021-09-10
Remove payload from logs and check for hash collision

zigaLuksic:


sentinelhub/download/client.py
```diff
@@ -144,6 +145,22 @@ def _is_download_required(self, request, response_path):
...
 
+    @staticmethod
+    def _check_cached_request_is_matching(request, request_path):
+        """ Ensures that the cached request matches the current one. Serves as protection against hash collisions
+        """
+        cached_request_info = read_data(request_path, MimeType.JSON)
+        current_request_info = request.get_request_params(include_metadata=True)
+        # Timestamps are allowed to differ
+        del cached_request_info['timestamp']
+        del current_request_info['timestamp']
```
AleksMat, line 156: We calculate hash with `get_request_params(include_metadata=False)`, therefore we should also here use `include_metadata=False`. Then you won't even have to delete timestamps.

The reason we don't want to compare headers (i.e. the other thing considered as metadata) is because they typically don't or shouldn't affect the response and because headers also contain a user agent tag with the package version. If someone upgrades their package version we don't want that caches become invalid only because of that.

PR #211, opened 2021-10-05
Improve tests and switch to f-strings

zigaLuksic:


tests/test_download.py
```diff
@@ -69,6 +69,7 @@ def download_request_fixture(output_folder):
     )
 
 
+@pytest.mark.sh_integration
```
AleksMat, line 72: In this module I wouldn't mark any of the tests with `sh_integration`. This roda.sentinel-hub.com is actually just a http access to info about the content of Sentinel S3 buckets. It is completely public.

AleksMat (commented):
I really like the solution in `testing_utils`

A few notes I couldn't write in the code:

- In `test_areas.py` a test case with `TileSplitter` is actually calling SH WFS so it should be marked with `sh_integration`.
- In `test_data_request.py` the test `test_saving_responses` should actually be marked with `aws_integration` while the other two are correctly marked as `sh_integration`.
- In `test_ogc.py` I noticed that test functions have weird names. Function `test_wfs` should probably be named `test_ogc` and `test_results` should probably be named `test_wfs`.

AleksMat (approved):

PR #222, opened 2021-11-11
TileSplitter update

AleksMat:
The main change is that now `TileSplitter` uses Catalog API to obtain grid geometries and that it supports all data collections, which are supported by the Catalog. Because of that I had to make a few code-breaking changes in class parameters and responses obtained from `get_info_list` method.

Additional changes in this PR:
- added `SHConfig.copy` method,
- updates according to `shapely` `1.8.0`
  * for the new `shapely` versions we now use `unary_union` which actually seems to be more consistent with `cascaded_union` in the older versions,
- updated notebook examples.


sentinelhub/config.py
```diff
@@ -215,6 +215,11 @@ def save(self):
         if is_changed:
             self._instance.save_configuration()
 
+    def copy(self):
```
zigaLuksic, line 218: Why not just use `copy.copy`?

zigaLuksic (approved):

sentinelhub/config.py
AleksMat: Due to a complex internal structure of `SHConfig` I'm always worried that at some point in the future `copy.copy` won't work correctly. Now that we have a method, which is unit-tests, we can just use this.

PR #223, opened 2021-12-06
Fixed metabands and their units

AleksMat:
I fixed a couple of issues regarding metabands and their units. I also double-checked that everything else is ok.


zigaLuksic (approved):
Looks good

PR #228, opened 2022-01-05
MimeType enums needed for eo-learn and an extra method

AleksMat:
Updates and improvements needed for https://github.com/sentinel-hub/eo-learn/pull/347#discussion_r778714451

### DIFF
--- a/sentinelhub/constants.py
+++ b/sentinelhub/constants.py
@@ -323,6 +323,11 @@ class MimeType(Enum):
     TAR = 'tar'
     RAW = 'raw'
     SAFE = 'safe'
+    PICKLE = 'pkl'
+    NPY = 'npy'
+    GPKG = 'gpkg'
+    GEOJSON = 'geojson'
+    GZIP = 'gz'
 
     @property
     def extension(self):
@@ -398,10 +403,6 @@ class MimeType(Enum):
         :return: String describing the file format
         :rtype: str
         """
-        if self is MimeType.TAR:
-            return 'application/x-tar'
-        if self is MimeType.JSON:
-            return 'application/json'
         if self is MimeType.JP2:
             return 'image/jpeg2000'
         if self is MimeType.XML:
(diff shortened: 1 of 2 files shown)


sentinelhub/constants.py
```diff
@@ -410,6 +411,14 @@ def get_string(self):
             return self.value
         return mimetypes.types_map['.' + self.value]
 
+    def is_extension_of(self, path: str) -> bool:
```
batic, line 414: Weird name, couldn't understand what the method does without context or reading docs. But I also don't have a better one. It also inverts the functionality: one is used to the check `path.endswith(mimetype.extension())` (is this file of such type), but not the other way around (is this type correct for this file)... 

On a quick glance I only see this used in tests. If that is so, suggest removing it and rather rewrite test to follow the typical logic (is this file of such type).

AleksMat: It is being used at 2 places in https://github.com/sentinel-hub/eo-learn/pull/348. At first I though I will use it in even more places but then there was no need.

Note that checking `path.endswith(mime_type.extension)` would be wrong. E.g. `'myfile.geojson'.endswith(MimeType.JSON.extension)` would return `True` because the `extension` property in `MimeType` class doesn't add `.`. So the correct way would be `path.endswith('.' + mime_type.extension)` which is a bit too verbose and deserves its own method.

The name is indeed a bit weird, would it perhaps be better `is_extension_of_path`?

zigaLuksic: How about `matches_file_format`?

AleksMat: Perhaps `matches_extension`? It is checking only a path that has a file extension in its name, not actual file format of a file.

batic: 👍 for `matches_extension`

AleksMat: Renamed.

PR #229, opened 2022-01-19
Support AWS session tokens

theirix:
Added a config parameter `aws_session_token` that allows configuring libraries with `SHConfig` instances with temporary IAM credentials that are usually used at EC2 machines (access id, key and STS token are all required to initialize `boto3`).


AleksMat (approved):
Hi @theirix, thank you for this contribution. It will be merged into `develop` branch and during the next version release to `master`.

theirix:
That's great, thank you!

PR #233, opened 2022-02-08
Remove dependency on wheel

adamchainz:
This package does not depend on `wheel` being installed to run - there are no imports of `wheel` in the code. Indeed it's not even distributed as a wheel :)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-wheel
 requests>=2.5.0
 click
 numpy


AleksMat:
Hi @adamchainz,

I agree with removing the `wheel` dependency. In the past, I think we had some issues if `wheel` wasn't installed because some environments tried to create a wheel from `sentinelhub-py` package. But having `wheel` package under package dependencies is not correct.

We'll merge this change to `develop` and at the next package release to `master`.

Thanks for pointing this out. :+1:

PR #237, opened 2022-02-23
Add black and isort to sh-py

mlubej:
Changes:
- added black and isort to requirements-dev.txt
- added pyproject.toml
- ran black and isort on all files


mlubej:
I'm not sure if this is something that's desired for the sh-py package. I noticed `eo-learn` and `eo-grow` having this so I added it also here.

Feel free to use this MR also for implementing gitlab actions.

mlubej:
I see some checks fail. I can fix them if it's desired to have this MR.

zigaLuksic:
I think we'll do this switch sooner or later, since we adopted it in eo-learn already. @AleksMat what is you opinion?

AleksMat:
Yes, pylint is failing because of the black formatting that was applied. I see that in some cases now lines are more than 120 characters which shouldn't happen. For some other things we can just increase the limits in `.pylintrc` for e.g. lines per file and number of matching lines for duplication.

Since these issues are specific to this PR they should be fixed here.

zigaLuksic:
> Yes, pylint is failing because of the black formatting that was applied. I see that in some cases now lines are more than 120 characters which shouldn't happen. For some other things we can just increase the limits in `.pylintrc` for e.g. lines per file and number of matching lines for duplication.
> 
> Since these issues are specific to this PR they should be fixed here.

It is very likely docstrings, those have to be fixed by hand since blacks string formatting is still a work in progress

PR #239, opened 2022-02-23
Update plotting function in notebook

mlubej:
I updated the plotting function in the notebook for batch. The function disolves the tiles because there are issues if many are present, then it plots each multipolygon for each status.

Also the legend has been fixed and set to something more intuitive.

Before:

image

After:

image


AleksMat (changes requested):
I see that besides updating the code in a couple of cells you also unnecessary updated all plotted images which technically don't change. This is just blowing up git history and makes it hard to see what actually changed in this PR. Therefore I suggest the following: 

- take the previous version of this notebook,
- replace only the content of cells that you actually modified and commit,
- you don't have to rerun everything again but if you decide to do just don't commit the new images (which will be virtually the same as the old ones).

Otherwise, the changes in code look good. :+1:

mlubej:
> I see that besides updating the code in a couple of cells you also unnecessary updated all plotted images which technically don't change. This is just blowing up git history and makes it hard to see what actually changed in this PR. 

Hmm, weird. I thought I did exactly that, at least these were the only differences shown in my IDE. Will check again.

AleksMat:
Ah, I see now that it actually just removed a `\n` character in each of those lines :facepalm:. Ok, I'll stop complicating about this.

mlubej:
It seems the culprit was the `vscode` notebook renderer, adding these changes in the background or something. Now the diff is clean when produced with the original jupyter notebook

AleksMat (approved):

PR #240, opened 2022-03-01
Modifications to support S2 v04.00

sebastianclarke:
Support the v04.00 processing baseline from ESA and allow the downloaded products to be used with FMask.


AleksMat:
Thanks for this nice contribution!

We'll have to add a few tests to make sure that we correctly support all new versions of .SAFE format. After that these changes will be released in the next version of the package.

Issue #231, opened 2022-01-27
[BUG] SAFE download format doesn't fetch QI_DATA files with Sentinel 2 L1C v04.00


anthony-scarth:
**Describe the bug**
Sentinel 2 L1C Product Version 04.00 has some significant changes to the contents of the QI_DATA folder and the sentinelhub downloader no longer fetches them correctly (when downloading in SAFE format - other downloads untested)

**_The main issue is that the contents of the QI folder which used to be GML and are now JP2_**
There have also been some name changes, reducing the total number of files available

A comparison of the contents can be seen with:
```
aws s3 ls s3://sentinel-s2-l1c/tiles/30/U/VD/2022/1/27/0/qi/ --request-payer requester  # new format
aws s3 ls s3://sentinel-s2-l1c/tiles/30/U/VD/2022/1/24/0/qi/ --request-payer requester  # old format
```

**To Reproduce**

Run the following python lines and inspect the downloaded data
```
from sentinelhub.data_request import download_safe_format
download_safe_format(
            product_id="S2B_MSIL1C_20220127T113309_N0400_R080_T30UVD_20220127T121733",
            folder="/tmp",
            redownload=True,
        )
```
The new JP2 files will be missing from the downloaded QI_DATA folder at:
/tmp/S2B_MSIL1C_20220127T113309_N0400_R080_T30UVD_20220127T121733.SAFE/GRANULE/L1C_T30UVD_A025560_20220127T113435/QI_DATA

**Expected behavior**

The downloader should fetch the new versions of these files
e.g.
DETFOO_B01.jp2 and QUALIT_B01.jp2
and structure them in the QI_DATA subfolder of the SAFE format

NOTE: they will need to be renamed upon download to MSK_DETFOO_B01.jp2 and MSK_QUALIT_B01.jp2 to match ESA's format!

e.g. Within 
S2B_MSIL1C_20220127T113309_N0400_R080_T30UVD_20220127T121733.SAFE/GRANULE/L1C_T30UVD_A025560_20220127T113435/QI_DATA

Should be:
./MSK_DETFOO_B01.jp2
./MSK_QUALIT_B01.jp2
./MSK_DETFOO_B02.jp2
etc...

**Environment**

sentinelhub 3.4.2

**Stack trace**

Output will likely print these warnings for all "missing" files (but otherwise complete the download as normal):
```
[...]

PR #243, opened 2022-03-20
Updates of AWS .SAFE format functionalities

AleksMat:
This is a continuation of the work started in https://github.com/sentinel-hub/sentinelhub-py/pull/240

This PR has a lot of changes but they are basically grouped in 3 parts according to the 3 commits:

1. First, I added tests for all new .SAFE baseline version up to the newest 04.00 and made a few changes in the code to make sure that the correct structure is always reconstructed (solves https://github.com/sentinel-hub/sentinelhub-py/issues/231). I also fixed an issue from the past where we didn't reconstruct datastrip reports which were not available on S3 at the time but now are.
2. I refactored the tests for .SAFE format which have grown quite a bit.
3. I refactored some parts of the code in `aws_safe.py` so that it is now a bit better structured. 

Note that the code of AWS functionalities in the package is still quite convoluted but any larger refactoring would take much more time and might not be easy to achieve. After all, the naming rules on S3 and of .SAFE format are also convoluted because of so many changes between different versions.


sentinelhub/aws_safe.py
```diff
@@ -33,34 +33,66 @@ def get_safe_struct(self):
...
+        """Builds a datastrip subfolder structure of .SAFE format."""
+        datastrip_safe = {}
         datastrip_list = self.get_datastrip_list()
         for datastrip_folder, datastrip_url in datastrip_list:
-            safe[main_folder][AwsConstants.DATASTRIP][datastrip_folder] = {}
-            safe[main_folder][AwsConstants.DATASTRIP][datastrip_folder][AwsConstants.QI_DATA] = {}
-            # S-2 L1C reports are on AWS only stored with tiles and without RADIOMETRIC_QUALITY
-            if self.has_reports() and self.data_collection is DataCollection.SENTINEL2_L2A:
+            datastrip_safe[datastrip_folder] = {}
+            datastrip_safe[datastrip_folder][AwsConstants.QI_DATA] = {}
```
zigaLuksic, line 66: why not `datastrip_safe[datastrip_folder] = {AwsConstants.QI_DATA: {}}`?

sentinelhub/aws_safe.py
```diff
@@ -33,34 +33,66 @@ def get_safe_struct(self):
...
                 for metafile in AwsConstants.QUALITY_REPORTS:
                     metafile_name = self.add_file_extension(metafile)
-                    safe[main_folder][AwsConstants.DATASTRIP][datastrip_folder][AwsConstants.QI_DATA][
+
+                    # S-2 L1C reports are available on S3 under modified names
+                    metafile_s3_name = (
+                        metafile_name.replace(".xml", "_report.xml")
+                        if self.data_collection is DataCollection.SENTINEL2_L1C
+                        else metafile_name
+                    )
```
zigaLuksic, line 77: This is actually one line longer than using a regular `if else` structure

sentinelhub/aws_safe.py
```diff
@@ -33,34 +33,66 @@ def get_safe_struct(self):
...
+
+                    datastrip_safe[datastrip_folder][AwsConstants.QI_DATA][
                         metafile_name
-                    ] = f"{datastrip_url}/qi/{metafile_name}"
+                    ] = f"{datastrip_url}/qi/{metafile_s3_name}"
```
zigaLuksic, line 81: gotta say i'm not a fan of these linebreaks :thinking:

sentinelhub/aws_safe.py
```diff
@@ -33,34 +33,66 @@ def get_safe_struct(self):
...
 
             data_strip_name = self.get_datastrip_metadata_name(datastrip_folder)
-            safe[main_folder][AwsConstants.DATASTRIP][datastrip_folder][
+            datastrip_safe[datastrip_folder][
                 data_strip_name
             ] = f"{datastrip_url}/{self.add_file_extension(AwsConstants.METADATA)}"
```
zigaLuksic, line 86: same here

sentinelhub/aws_safe.py
```diff
@@ -71,25 +103,11 @@ def get_safe_struct(self):
...
-
-        safe[main_folder][self.get_product_metadata_name()] = self.get_url(AwsConstants.METADATA)
-        safe[main_folder]["INSPIRE.xml"] = self.get_url(AwsConstants.INSPIRE)
-        safe[main_folder][self.add_file_extension(AwsConstants.MANIFEST)] = self.get_url(AwsConstants.MANIFEST)
 
-        if self.is_early_compact_l2a():
-            safe[main_folder]["L2A_Manifest.xml"] = self.get_url(AwsConstants.L2A_MANIFEST)
-            safe[main_folder][self.get_report_name()] = self.get_url(AwsConstants.REPORT)
+                for tile_name, safe_struct in tile_struct.items():
+                    granule_safe[tile_name] = safe_struct
```
zigaLuksic, line 108: i think this can be `granule_safe.update(tile_struct)`

sentinelhub/aws_safe.py
```diff
@@ -195,75 +213,88 @@ def get_safe_struct(self):
...
-            safe[main_folder][AwsConstants.AUX_DATA][self.get_aux_data_name()] = self.get_url(ecmwft_file)
-        # Old products also have DEM and MSI in aux folder
+            aux_safe[self.get_aux_data_name()] = self.get_url(ecmwft_file)
+
+            if self.baseline >= "04.00":
+                camsfo_file = (
+                    AwsConstants.CAMSFO
+                    if self.data_collection is DataCollection.SENTINEL2_L1C
+                    else AwsConstants.AUX_CAMSFO
+                )
```
zigaLuksic, line 246: a regular `if else` would be shorter

sentinelhub/aws_safe.py
```diff
@@ -272,26 +303,41 @@ def get_safe_struct(self):
...
-        preview_type = (
-            "L2A"
-            if (
-                self.data_collection is DataCollection.SENTINEL2_L2A
-                and (self.baseline >= "02.07" or self.baseline == "00.01")
-            )
-            else "L1C"
+        is_newer_l2a_version = self.data_collection is DataCollection.SENTINEL2_L2A and (
+            self.baseline >= "02.07" or self.baseline == "00.01"
         )
```
zigaLuksic, line 313: Does it format it any nicer if you put the parenthesis around the whole expression? Something like
```python
x = (
    exp1
    and exp2
)
```

zigaLuksic (commented):
Great job aside from some minor style mishaps.
Really glad I didn't have to do this :D

sentinelhub/aws_safe.py
AleksMat: fixed

sentinelhub/aws_safe.py
AleksMat: changed

sentinelhub/aws_safe.py
AleksMat: fixed

sentinelhub/aws_safe.py
AleksMat: fixed

sentinelhub/aws_safe.py
AleksMat: nice catch

sentinelhub/aws_safe.py
AleksMat: In this case I would actually leave it as it is. Somehow this way seems more readable to me.

sentinelhub/aws_safe.py
AleksMat: Sadly no - black just automatically reverts back to the previous formatting. I would leave it like it is.

zigaLuksic: sad :( but yeah, useless to fight against an autoformatter

Issue #230, opened 2022-01-26
[BUG] Limit maximal pillow version introduced a cve to all downstream users


lingtong-lu:
**Describe the bug**

This [specific commit](https://github.com/sentinel-hub/sentinelhub-py/commit/1c7818a058e96e6bbf5a4e2dd9723eac18683930) limits the maximal pillow version to be **8.4.0**, which introduced a cve to all downstream users

CVE:
* https://github.com/advisories/GHSA-8vj2-vxx3-667w
* https://github.com/advisories/GHSA-xrcv-f9gm-v42c
* https://github.com/advisories/GHSA-pw3c-h7wp-cvhx

**To Reproduce**

Limit maximal pillow version introduced a cve to all downstream users

**Expected behavior**

Limit maximal pillow version introduced a cve to all downstream users

**Environment**

Python 3.8

**Stack trace or screenshots**

```
  pillow<=8.4.0,>=8.3.2 (from sentinelhub==3.4.3->s2cloudless==1.5.0->-r /var/folders/gg/d6wg3b1x5v73sh76m82fjcl00000gn/T/pipenv244mtkg9requirements/pipenv-xq7z0lxo-constraints.txt (line 7))
```

**Desktop (please complete the following information):**

N/A

**Additional context**

batic:
Thanks for bringing this to our attention. 

PIL is being used when reading jpeg2000 files, and unfortunately newer versions don't read correctly Sentinel-2 (15-bit) images, hence the version was fixed. We will have a look and see how to best resolve the Pillow version while keeping the functionality.

AleksMat:
Solved in the newly released version `3.5.0`.

PR #249, opened 2022-03-24
Handling new Pillow version with rasterio alternative

AleksMat:
Solves https://github.com/sentinel-hub/sentinelhub-py/issues/230

- For decoding JPEG images we accept changes brought by new Pillow version.
- Decoding 15-bit JP2 images cannot be done with the new Pillow version. In such case users will now get a warning and be suggested to install `rasterio`. If `rasterio` is installed then `sentinelhub-py` will use it to decode JP2 images by default.

Additional improvements in this PR:
- better handling of some warnings,
- updated deprecated function and parameter from `tifffile` package,
- moved pytest and code coverage configurations into `pyproject.toml` to resolve some warnings because of clashes of multiple configuration files.


zigaLuksic (approved):
Nicely done

PR #257, opened 2022-04-07
Feat/tokeninfo

batic:
This PR adds a method to `SentinelHubSession` that parses the token to obtain tokeninfo. 

Disclaimer: Same information can be obtained at `tokeninfo` endpoint, but decoding it saves a request.

### DIFF
--- a/tests/download/test_session.py
+++ b/tests/download/test_session.py
@@ -6,8 +6,6 @@ from sentinelhub import SentinelHubSession
 @pytest.fixture(name="session", scope="module")
 def session_fixture():
     session = SentinelHubSession()
-    # pre-fetch token
-    session.token
     return session
 
 
--- a/tests/download/test_session.py
+++ b/tests/download/test_session.py
@@ -3,11 +3,16 @@ import pytest
 from sentinelhub import SentinelHubSession
 
 
-@pytest.mark.sh_integration
-def test_session():
-
+@pytest.fixture(name="session")
+def session_fixture():
     session = SentinelHubSession()
+    # pre-fetch token
+    session.token
+    return session
 
+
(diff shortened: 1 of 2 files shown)


tests/download/test_session.py
```diff
@@ -3,11 +3,16 @@
...
 
 
-@pytest.mark.sh_integration
-def test_session():
-
+@pytest.fixture(name="session")
+def session_fixture():
     session = SentinelHubSession()
+    # pre-fetch token
+    session.token
```
AleksMat, line 10: It might not be the best practice to interact with SH services already in the fixture but then only tests are marked with `@pytest.mark.sh_integration`. Note that by default the fixture will get called once by every test and not just once in total. So this doesn't even reduce the number of new sessions created. Instead I propose a fixture like this:

```Python
@pytest.fixture(name="session", scope="module")
def session_fixture(config):
    return SentinelHubSession(config=config)
```

- By adding this scope the fixture will get called only once per entire module. So the same object will be passed to both tests and only 1 session will be created.
- Initializing session with config object that is obtained from `conftest.py`. This way we can propagate the same credentials throughout all tests.

batic: Yeah, as soon as I pushed this crossed my mind. Thanks for this. I like the `scope`, and will use it if it is ok with you. 

Putting this to `conftest.py` would make sense if we can reuse session in other tests and fixtures, e.g. with download clients etc.

AleksMat: For now I think it's ok if we leave session creation only in this module. That is because it is a bit tricky to create a session and then tell a download client to cache it for its uses. Maybe some time in the future if this will actually become an issue.

PR #258, opened 2022-04-07
Feat/byoc bands

batic:
This PR fixes up proper encoding/decoding of `additional_data` into `ByocCollection`, and adds a dataclass for `ByocCollectionBand`, which allows user to specify in advance the bands of their BYOC collection.

### DIFF
--- a/sentinelhub/api/byoc.py
+++ b/sentinelhub/api/byoc.py
@@ -36,7 +36,7 @@ class ByocCollectionAdditionalData:
     """Dataclass to hold BYOC collection additional data"""
 
     other_data: CatchAll
-    bands: Dict[str, ByocCollectionBand] = None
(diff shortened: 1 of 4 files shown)


sentinelhub/api/byoc.py
```diff
@@ -17,17 +17,32 @@
...
+
+
 @dataclass_json(letter_case=LetterCase.CAMEL, undefined=Undefined.INCLUDE)
 @dataclass
 class ByocCollectionAdditionalData:
     """Dataclass to hold BYOC collection additional data"""
 
     other_data: CatchAll
-    bands: Optional[dict] = None
+    bands: Dict[str, ByocCollectionBand] = None
```
AleksMat, line 39: Because the default value is `None`, should this have a type `Optional[Dict[str, ByocCollectionBand]]`?

PR #259, opened 2022-04-15
Add mosaicking and resampling enums 

zigaLuksic:


sentinelhub/constants.py
```diff
@@ -54,6 +54,22 @@ class ServiceType(Enum):
     PROCESSING_API = "processing"
 
 
+class ResamplingType(Enum):
```
AleksMat, line 57: While checking if service is case-sensitive about these parameters I discovered this:

```
"possibleValues":["BILINEAR","BICUBIC","LANCZOS","BOX","NEAREST"]
```

It turns out the 2 missing options work but they are not written in SH documentation. I'm not completely sure but I think it's better that we include them in the enum.

sentinelhub/api/ogc.py
```diff
@@ -416,7 +416,13 @@ def _get_common_url_parameters(request):
                 if geometry.crs is CRS.WGS84:
                     geometry = geometry.reverse()
 
-                params[CustomUrlParam.GEOMETRY.value] = geometry.wkt
+                custom_params[CustomUrlParam.GEOMETRY] = geometry.wkt
+
+            for resampling in (CustomUrlParam.DOWNSAMPLING, CustomUrlParam.UPSAMPLING):
+                if resampling in custom_params:
+                    custom_params[resampling] = ResamplingType(custom_params[resampling]).value
```
AleksMat, line 423: While all newer SH APIs are case-sensitive about parameter values, the old OGC is not. Therefore, it worked even with a value `"bicubic"` or even `"BiCuBiC"`. But now with this enum parsing it won't work anymore.

I'm wondering if we should add an extra check here that would convert a string to uppercase. Another option would be to even add a simple metaclass for `ResamplingType` where converting a string to uppercase would become a part of casting to enum. We already know how to do this because we do it for `CRS` enum.

sentinelhub/constants.py
batic: Please keep the possible values aligned with docs. Anything else is not officially supported, so we (sh-py) should not be exposing it.

sentinelhub/api/ogc.py
zigaLuksic: I would prefer to follow the wisdom "if you think you need a meta-class, you probably don't". I think it's best to solve this via function (or method) `parse_resampling_type` or by in-lining the parsing in the OGC and leaving it case sensitive elsewhere.

sentinelhub/constants.py
AleksMat: Ok, then let's just keep the official ones.

PR #263, opened 2022-04-25
More type annotations

zigaLuksic:
Since `py.typed` is included we can now start adding type annotations. This won't be easy so we should probably start with the low-hanging fruit and work our way to (possibly code-breaking) large changes.

I believe we should do this change gradually and in many small MRs so that the changes are actually looked at.

I tried to change the code itself as little as possible. In some places the precise types are rather verbose (a tuple of 4 tuples and the likes...)

### DIFF
--- a/sentinelhub/commands.py
+++ b/sentinelhub/commands.py
@@ -2,14 +2,18 @@
 Module that implements command line interface for the package
 """
 
+from typing import Any, Callable, TypeVar
+
 import click
 
 from .config import SHConfig
 from .download import DownloadClient, DownloadRequest
 
+FC = TypeVar("FC", bound=Callable[..., Any])
+
 
 @click.command()
-def main_help():
+def main_help() -> None:
     """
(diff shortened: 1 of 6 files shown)


sentinelhub/commands.py
```diff
@@ -2,14 +2,18 @@
...
 """
 
+from typing import Any, Callable, TypeVar
+
 import click
 
 from .config import SHConfig
 from .download import DownloadClient, DownloadRequest
 
+FC = TypeVar("FC", bound=Callable[..., Any])
```
AleksMat, line 12: Should we make the name `FC` a bit more verbose to be clear what it stands for? Maybe `_FunctionType`?

zigaLuksic: I used `FC` because that is what `click` uses. We can be more verbose if needed.

AleksMat: Hm, in that case let's leave it as it is.

PR #264, opened 2022-04-25
Improved parsing of WGS84 in CRS class

AleksMat:
We define WGS 84 in pyproj with reversed coordinates and transforming such object to EPSG code works:

```Python
from pyproj import CRS

CRS("+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs").to_epsg()

>> 4326
```

But, as far as I understand, during saving to Geopackage and loading from Geopackage the following happens with the object:

```Python
from pyproj import CRS

wgs84 = CRS("+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs")
new_wgs84 = CRS(wgs84.to_wkt())
new_wgs84.to_epsg()

>> None
```

Because EPSG code somehow can't be reconstructed this becomes an issue in `sentinelhub.CRS` parsing. The fix relies on the following:

```Python
from pyproj import CRS

wgs84 = CRS("+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs")
new_wgs84 = CRS(wgs84.to_wkt())
wgs84 == new_wgs84

>> True
```

Basically, we start relying more on the equality function than on `to_epsg`.

### DIFF
--- a/sentinelhub/constants.py
+++ b/sentinelhub/constants.py
@@ -129,6 +129,9 @@ class CRSMeta(EnumMeta):
             if epsg_code is not None:
                 return str(epsg_code)
 
+            if value == CRS.WGS84.pyproj_crs():
+                return "4326"
+
             error_message = f"Failed to determine an EPSG code of the given CRS:\n{repr(value)}"
             maybe_epsg = value.to_epsg(min_confidence=0)
             if maybe_epsg is not None:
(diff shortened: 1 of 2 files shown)


zigaLuksic (approved):
should this be brought to the attention of the `pyproj` team?

PR #265, opened 2022-04-28
Fixed issue of passing SHConfig to Ray

AleksMat:
The problem was that when an instance of `SHConfig` would be passed to a Ray worker it would lose `self._instance` because that is defined only as a class attribute. However, `self._instance` was being called in many methods, even in `__repr__`. So calling `repr(config)` on Ray worker would fail.

The PR only contains a fix and an extra test for this.More improvements will be done in near future where we will also aim to get autocomplete for config parameters.

This PR also removes and limits some overrides of `__getatt__` and `__getitem__` methods and also puts adds Ray to development requirements for Python version `<3.10`.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,4 +9,4 @@ isort
 pillow>=9.0.0
 rasterio;python_version<"3.10"
 click>=8.0.0
-ray
+ray[default]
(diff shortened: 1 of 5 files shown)


zigaLuksic:
I believe the Config class is long overdue for an overhaul.

PR #266, opened 2022-05-04
Types in `areas.py`

zigaLuksic:
The MR comes in three commits, and I suggest it is best reviewed this way.
1. Change of `_make_split` so that it is now part of the super class, where all subclasses first set any attributes and then call `super().__init__`. This is a *90% there* improvement, because a 100% would need metaclasses and we don't want that. The method also now returns lists, instead of modifying attributes (risky)
2. Add type annotations. `UtmSplitter`s need a `type: ignore` because they redefine the interface of `get_bbox_list`.
3. Some more refactorings.

There are three similar functions for `_parse_split_parameters` that could be unified with some advanced `overload` use, I can try and do that if you feel it is necessary

### DIFF
--- a/sentinelhub/areas.py
+++ b/sentinelhub/areas.py
@@ -6,7 +6,7 @@ import json
 import math
 import os
 from abc import ABCMeta, abstractmethod
-from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
+from typing import Any, Dict, Iterable, List, Optional, Tuple, TypeVar, Union, cast
 
 import shapely
 import shapely.geometry
@@ -22,6 +22,8 @@ from .constants import CRS
 from .data_collections import DataCollection
 from .geo_utils import transform_point
 from .geometry import BBox, BBoxCollection, Geometry, _BaseGeometry
+
+T = TypeVar("T", float, int)
 
 
 class AreaSplitter(metaclass=ABCMeta):
@@ -195,21 +197,8 @@ class BBoxSplitter(AreaSplitter):
         :param reduce_bbox_sizes: If `True` it will reduce the sizes of bounding boxes so that they will tightly fit
             the given area geometry from `shape_list`.
         """
-        self.split_shape = self._parse_split_parameters(split_shape)
+        self.split_shape = _parse_to_pair(split_shape, allowed_types=(int,), param_name="split_shape")
         super().__init__(shape_list, crs, **kwargs)
-
-    @staticmethod
-    def _parse_split_parameters(split_parameter: Union[int, Tuple[int, int]]) -> Tuple[int, int]:
-        """Parses the parameters defining the splitting of the BBox."""
-        if isinstance(split_parameter, int):
-            return split_parameter, split_parameter
-
-        if isinstance(split_parameter, (tuple, list)) and len(split_parameter) == 2:
-            split_x, split_y = split_parameter
-            if isinstance(split_x, int) and isinstance(split_y, int):
-                return split_x, split_y
-
-        raise ValueError(f"Split parameter must be an integer or a pair of integers but {split_parameter} was given")
 
     def _make_split(self) -> Tuple[List[BBox], List[Dict[str, object]]]:
         columns, rows = self.split_shape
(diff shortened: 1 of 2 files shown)


sentinelhub/areas.py
```diff
@@ -230,39 +193,51 @@ class BBoxSplitter(AreaSplitter):
...
         :param reduce_bbox_sizes: If `True` it will reduce the sizes of bounding boxes so that they will tightly fit
             the given area geometry from `shape_list`.
-        :type reduce_bbox_sizes: bool
         """
+        self.split_shape = self._parse_split_parameters(split_shape)
         super().__init__(shape_list, crs, **kwargs)
 
-        self.split_shape = self._parse_split_parameters(split_shape)
+    @staticmethod
+    def _parse_split_parameters(split_parameter: Union[int, Tuple[int, int]]) -> Tuple[int, int]:
```
AleksMat, line 216: Hm, it would be nice to unify the 3 almost the same methods but its not a high priority. I'm wondering if you can do that with TypeVars, something like:

```Python
def _parse_split_parameters(parameter: Union[_Type, Tuple[_Type, _Type]]) -> Tuple[_Type, _Type]:
```

Although this type is then also used for checking in the method, so the method might need a parameter to specify which type you want. Then the new type parameter would have to be typed with the same type var and I'm not sure if that is possible. :sweat_smile: 

So if you don't find a nice way you can just leave it.

sentinelhub/areas.py
```diff
@@ -361,42 +335,41 @@ class TileSplitter(AreaSplitter):
...
         :param kwargs: Parameters that are propagated to the base `AreaSplitter` class
         """
-        super().__init__(shape_list, crs, **kwargs)
-
         self.time_interval = time_interval
         self.tile_split_shape = tile_split_shape
         self.data_collection = data_collection
 
-        config = config or SHConfig()
+        sh_config = config or SHConfig()
```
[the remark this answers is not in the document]
zigaLuksic: the issue here is that i'm overriding a `Optional[Config]` variable with a `Config` value, but this is not reflected in the type unless I use a different name. This is related to the "please dont re-use names" issue.

sentinelhub/areas.py
zigaLuksic: I can try to conjure up something with `override`. But I would probably need `Literal`, not entirely sure. TypeVars are probably not strong enough here as you noted yourself.

sentinelhub/areas.py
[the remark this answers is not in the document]
AleksMat: Ah, I see. Then let's leave this.

sentinelhub/areas.py
```diff
@@ -5,51 +5,57 @@
...
+    def _parse_shape_list(
+        shape_list: Iterable[Union[Polygon, MultiPolygon, _BaseGeometry]], crs: CRS
+    ) -> List[Union[Polygon, MultiPolygon]]:
         """Checks if the given list of shapes is in correct format and parses geometry objects
 
         :param shape_list: The parameter `shape_list` from class initialization
-        :type shape_list: list(shapely.geometry.multipolygon.MultiPolygon or shapely.geometry.polygon.Polygon)
         :raises: ValueError
         """
         if not isinstance(shape_list, list):
```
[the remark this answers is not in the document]
zigaLuksic: after this the method can be inlined which i did :)

sentinelhub/areas.py
zigaLuksic: I am actually going to try and do this, it seems like `@overload` is pretty strong

zigaLuksic: Sadly, since we're not 3.8 and there is no neat supertype for float+int, I had to make some sacrifices.
Since the parameters are already typed, the functions type can be fairly non-complex, just needs a bit of generics.

Some confusion also follows from the fact that in [PEP 484](https://peps.python.org/pep-0484/#the-numeric-tower) it is stated that `float` is the proper annotation for `float | int` types, but `isinstance(1, float)` fails. Therefore the type validation was decoupled from the type signature of the function. Once we go to 3.8 we can adjust this via overloading, until then it is what it is.

AleksMat: Ok, looks good :+1:

PR #273, opened 2022-05-23
Add outstanding MSK_CLASSI_B00 file from processing baseline 04.00

sebastianclarke:
These changes are required to enable the fetched products to support BioPhysical processing in SNAP.


AleksMat:
I agree it makes sense to add this constant which can then be used by users. Otherwise the current code for creating the .SAFE structure already includes `qi/MSK_CLASSI_B00`, it is just generated on-the-fly.

Merging this PR to develop and it will be released in the next package version `3.6.0`. Thanks for the contribution!

PR #276, opened 2022-05-24
Add types to sentinelhub.download

zigaLuksic:
Sadly not all of the methods type easily, due to over-generalisation. I feel like this is not the time to refactor the code too much, so I mostly added some additional guards.

### DIFF
--- a/sentinelhub/api/base.py
+++ b/sentinelhub/api/base.py
@@ -6,7 +6,7 @@ import warnings
 from abc import ABCMeta
 from dataclasses import dataclass, field
 from datetime import datetime
-from typing import Any, Dict, Iterable, Optional, Union
+from typing import Any, Dict, Iterable, Optional, Union, cast
 from urllib.parse import urlencode
 
 from dataclasses_json import CatchAll, LetterCase, Undefined
@@ -77,6 +77,7 @@ class SentinelHubFeatureIterator(FeatureIterator[Dict[str, Any]]):
         url = f"{self.url}?{urlencode(params)}"
 
         results = self.client.get_json(url, use_session=True)
+        results = cast(Dict[str, Any], results)
 
         new_features = results.get("data")
         if new_features is None:
--- a/sentinelhub/api/batch.py
+++ b/sentinelhub/api/batch.py
@@ -177,10 +177,7 @@ class SentinelHubBatch(SentinelHubService):
         :return: A tiling grid definition
         """
         url = self._get_tiling_grids_url(grid_id)
-        json_response = self.client.get_json(url=url, use_session=True)
-        if not isinstance(json_response, dict):
-            raise RuntimeError(f"Response from {url} did not contain the expected information.")
-        return json_response
+        return self.client.get_json_dict(url=url, use_session=True)
 
     def iter_requests(
         self, user_id: Optional[str] = None, search: Optional[str] = None, sort: Optional[str] = None, **kwargs: Any
@@ -345,10 +342,7 @@ class SentinelHubBatch(SentinelHubService):
         """
         request_id = self._parse_request_id(batch_request)
         url = self._get_tiles_url(request_id, tile_id=tile_id)
-        json_response = self.client.get_json(url, use_session=True)
-        if not isinstance(json_response, dict):
-            raise RuntimeError(f"Response from {url} did not contain the expected information.")
-        return json_response
+        return self.client.get_json_dict(url, use_session=True)
 
     def reprocess_tile(self, batch_request: BatchRequestType, tile_id: Optional[int]) -> Json:
         """Reprocess a single failed tile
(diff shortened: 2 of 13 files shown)


sentinelhub/api/batch.py
```diff
@@ -175,7 +176,11 @@ def get_tiling_grid(self, grid_id: Union[int, str]) -> Dict[str, Any]:
         :param grid_id: An ID of a requested tiling grid
         :return: A tiling grid definition
         """
-        return self.client.get_json(self._get_tiling_grids_url(grid_id), use_session=True)
+        url = self._get_tiling_grids_url(grid_id)
+        json_response = self.client.get_json(url=url, use_session=True)
+        if not isinstance(json_response, dict):
+            raise RuntimeError(f"Response from {url} did not contain the expected information.")
+        return json_response
```
AleksMat, line 183: This part of the code now happens so many time that it could even become its own method on the common base class. E.g. `SentinelHubService.get_json_data()` or something like that.

sentinelhub/download/client.py
```diff
@@ -87,11 +106,11 @@ def download(self, download_requests, max_threads=None, decode_data=True, show_p
                 for future in as_completed(download_list):
                     data_list[future_order[future]] = self._process_download_future(future)
 
-        if is_single_request:
+        if isinstance(download_requests, DownloadRequest):
             return data_list[0]
         return data_list
 
-    def _process_download_future(self, future):
+    def _process_download_future(self, future: Future) -> Any:
```
AleksMat, line 113: I'm not sure if it's work complicating but this could even be written as:

```Python
def _process_download_future(self, future: Future[T]) -> Optional[T]:
```

Where `Future[T]` works ok for Python >=3.9 and for lower versions probably `typing-extensions` backport can handle it.

sentinelhub/download/client.py
```diff
@@ -135,8 +156,11 @@ def _single_download(self, request, decode_data):
 
     @retry_temporary_errors
     @fail_user_errors
-    def _execute_download(self, request):
+    def _execute_download(self, request: DownloadRequest) -> bytes:
         """A default way of executing a single download request"""
+        if request.url is None:
+            raise RuntimeError(f"Faulty request {request}, no URL specified.")
```
AleksMat, line 162: I would actually suggest raising `ValueError` here because this error could only happen if a user provides wrong input parameters. `RuntimeError` is more natural for cases where the reason for an error is e.g. that SH service returned an unexpected output such as in such case:

```Python
        if not isinstance(json_response, dict) or "data" not in json_response:
            raise RuntimeError(f"Response from {url} did not contain the expected information.")
```

sentinelhub/download/sentinelhub_client.py
```diff
@@ -87,16 +90,19 @@ def _execute_download(self, request):
...
             return thread_unsafe_function(*args, **kwargs)
 
         with self.lock:
             return thread_unsafe_function(*args, **kwargs)
 
-    def _do_download(self, request):
+    def _do_download(self, request: DownloadRequest) -> Any:
         """Runs the download"""
+        if request.url is None:
+            raise RuntimeError(f"Faulty request {request}, no URL specified.")
```
AleksMat, line 104: Same as above I suggest `ValueError`.

sentinelhub/download/sentinelhub_statistical_client.py
```diff
@@ -70,22 +73,26 @@ def _single_download(self, request, _):
...
         return None
 
-    def _download_per_interval(self, request, time_intervals):
+    def _download_per_interval(self, request: DownloadRequest, time_intervals: Dict[int, Any]) -> dict:
         """Download statistics per each time interval"""
         interval_requests = []
         for time_interval in time_intervals.values():
             interval_request = copy.deepcopy(request)
+            if interval_request.post_values is None or "aggregation" not in interval_request.post_values:
+                raise RuntimeError("Unable to configure request for retrying by interval.")
```
AleksMat, line 94: Also this depends on user's inputs so I suggest `ValueError`.

sentinelhub/api/base.py
```diff
@@ -77,6 +77,7 @@ def _fetch_features(self) -> Iterable[Dict[str, Any]]:
         url = f"{self.url}?{urlencode(params)}"
 
         results = self.client.get_json(url, use_session=True)
+        results = cast(Dict[str, Any], results)
```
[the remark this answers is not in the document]
zigaLuksic: I do that in the next MR where i'm finalizing types in API

sentinelhub/api/batch.py
zigaLuksic: Will also be done in the next MR since this pattern also repeats itself in the API part.

sentinelhub/download/client.py
zigaLuksic: I dont think it would be of much use, since the download itself returns anything

sentinelhub/download/client.py
zigaLuksic: Was on the verge regarding the kind of exceptions. Since we often configure things on the fly (though we shouldn't) I thought runtime wouldn't be worst choice. Will adjust the rest.

sentinelhub/api/batch.py
zigaLuksic: Actually I'll add it to this one, no reason to wait

PR #278, opened 2022-05-25
Make `sentinelhub-py` mypy compliant

zigaLuksic:
This MR adds the remaining types to AWS and geopedia subpackages.

I then cleaned the docstrings to remove :type: mentions.
Finally I added mypy to CI.


sentinelhub/api/statistical.py
```diff
@@ -36,20 +36,14 @@ def __init__(
...
-        :type aggregation: dict
         :param input_data: A list of input dictionary objects as described in the API reference. It can be generated
             with `input_data` method
-        :type input_data: List[dict or InputDataDict]
         :param bbox: A bounding box of the request
-        :type bbox: sentinelhub.BBox or None
         :param geometry: A geometry of the request
-        :type geometry: sentinelhub.Geometry or None
         :param calculations: Calculations part of the payload.
         :param calculations: dict
```
AleksMat, line 44: Uf, this line should also be removed. An old mistake guess. :see_no_evil:

PR #280, opened 2022-05-26
Session sharing

AleksMat:
The main additions are:

- utilities for session sharing when parallelizing with standard Python multiprocessing,
- a notebook explaining the session interface and session sharing procedure.

There are also a few minor other improvements.


sentinelhub/download/session.py
```diff
@@ -126,3 +134,125 @@ def _fetch_token(self, request: DownloadRequest) -> JsonDict:
...
+    def run(self) -> None:
+        """A running thread is running an infinite loop of sharing a token and waiting for token to expire. The loop
+        ends only when the thread is stopped."""
+        self._stop_event.clear()
+
+        while not self._stop_event.is_set():
+            token = self.session.token
+            self._share_token(token)
+
+            sleep_until_refresh_time = token["expires_at"] - time.time() - self._refresh_time
```
zigaLuksic, line 200: wouldn't this refresh the token JUST as it becomes expired?
does that mean there is a small possibility of the token being grabbed after it expired but before it is refreshed? Can we put a lock on the shared memory while refreshing (and a bit of a buffer to the time where that happens so we're not a microsecond too late)?

sentinelhub/download/session.py
```diff
@@ -126,3 +134,125 @@ def _fetch_token(self, request: DownloadRequest) -> JsonDict:
...
+        length of encoded token can vary a bit.
+        """
+        encoded_token = json.dumps(token).encode()
+
+        if self._is_memory_shared_event.is_set():
+            memory = SharedMemory(name=self.memory_name)
+        else:
+            memory = SharedMemory(
+                create=True,
+                size=len(encoded_token) + self._EXTRA_MEMORY_BYTES,
```
zigaLuksic, line 217: this one worries me a bit because it could break if we significantly change the token, but i'm not sure there is another way of doing it

sentinelhub/download/session.py
```diff
@@ -126,3 +134,125 @@ def _fetch_token(self, request: DownloadRequest) -> JsonDict:
...
+    """This utility function is meant to be used in combination with `SessionSharingThread`. It retrieves an
+    authentication token from the shared memory and returns it in an `SentinelHubSession` object.
+
+    :param memory_name: A unique name of the requested shared memory block from where to read the session. It should
+        match the one used in `SessionSharingThread`.
+    :return: An instance of `SentinelHubSession` that contains the shared token but is not self-refreshing.
+    """
+    memory = SharedMemory(name=memory_name)
+    try:
+        encoded_token = memory.buf.tobytes().rstrip(_NULL_MEMORY_VALUE)
```
zigaLuksic, line 253: it cannot happen that this would remove too much right?

sentinelhub/download/session.py
```diff
@@ -126,3 +134,125 @@ def _fetch_token(self, request: DownloadRequest) -> JsonDict:
...
+
+    .. code-block:: python
+
+        thread = SessionSharingThread(session)
+        thread.start()
+
+        # Run a parallelization process here
+        # Use collect_shared_session() to retrieve the session with other processes
+
+        thread.stop()
```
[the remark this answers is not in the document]
AleksMat: It would be useful. The only problem I see is that because `SessionSharingThread` inherits directly from `Thread` users would expect the same interface. The fact that they would now be able to use a thread object with `with` statement might be a bit confusing.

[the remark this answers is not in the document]
AleksMat: In fact, I'm thinking about renaming `stop` into `join` because that is actually the default method used for stopping a thread.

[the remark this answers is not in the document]
zigaLuksic: I would leave this as is, might come in handy, but I would create a separate object that uses the `with` interface and is the suggested method of use

sentinelhub/download/session.py
AleksMat: `token["expires_at"] - time.time()` is the time until the token expires. With additional ` - self._refresh_time` we get the time until session will start refreshing token. So as long a `self._refresh_time` is not `0` (by default it is `120`), there will be still enough time to use the token for download.

Note that the shared memory is updated with a single command. This way I'm sure shared memory implementation makes sure that read and write operations cannot happen at the same time.

sentinelhub/download/session.py
AleksMat: In practice token size changes maybe +-3 characters because `"expires_at"` can have different number of decimals. I put `100` characters spare space just in case. Unless they change something at SH service that it would be returning tokens of very varying sizes, the current extra memory should be enough.

sentinelhub/download/session.py
AleksMat: Nope because the last valid byte will always be `b"}"`.

sentinelhub/download/session.py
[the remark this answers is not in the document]
AleksMat: Hm, I'm not a fan of having yet another object for that. Note that `with` functionality can be simply obtained with:

```Python
try:
    # parallelization code
finally:
    thread.stop()
```

sentinelhub/download/session.py
zigaLuksic: ah, i see, we're waiting for the window to start, should be a good enough buffer yeah.

probably not even god knows what python considers to be an atomic operation :joy:  but the buffer is large enough so that I dont think we need to worry

sentinelhub/download/session.py
zigaLuksic: alright, lets leave it then

sentinelhub/download/session.py
zigaLuksic: alright

sentinelhub/download/session.py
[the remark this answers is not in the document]
zigaLuksic: Unless we add a new object for it I am against adding this functionality. No more *Swiss-army-knife* objects please

sentinelhub/download/session.py
```diff
@@ -126,3 +134,125 @@ def _fetch_token(self, request: DownloadRequest) -> JsonDict:
...
+    """
+
+    _EXTRA_MEMORY_BYTES = 100
+
+    def __init__(
+        self, session: SentinelHubSession, *args: Any, memory_name: str = _DEFAULT_SESSION_MEMORY_NAME, **kwargs: Any
```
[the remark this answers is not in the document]
AleksMat: Removed `*args`.

sentinelhub/download/session.py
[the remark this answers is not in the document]
AleksMat: Ok, now I renamed to `thread.join` which makes much more sense. I also added try-finally statement into documentation example so that users are aware of this.

examples/session_sharing.ipynb
```diff
@@ -0,0 +1,585 @@
...
+    "\n",
+    "serialized_token = json.dumps(session.token)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "dedicated-adrian",
+   "metadata": {},
+   "source": [
+    "The following code would deserialize authentication token a cache it into a download client of another Python process:"
```
zigaLuksic, line 328: typo in `token a cache it`

examples/session_sharing.ipynb
```diff
@@ -0,0 +1,585 @@
...
+    "handler.setFormatter(formatter)\n",
+    "\n",
+    "log_filter = SessionFilter()\n",
+    "handler.addFilter(log_filter)\n",
+    "\n",
+    "global_logger = logging.getLogger()\n",
+    "global_logger.setLevel(logging.DEBUG)\n",
+    "\n",
+    "for handler in global_logger.handlers:\n",
+    "    global_logger.removeHandler(handler)\n",
```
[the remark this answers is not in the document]
AleksMat: Yes, it removes the default handler because we have to set the default global logger to debug.

Actually, now I got an idea how to write this a bit differently. I'll try it out.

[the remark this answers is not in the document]
AleksMat: It worked! Logging setup code is now much simpler.

zigaLuksic (approved):

examples/session_sharing.ipynb
```diff
@@ -0,0 +1,585 @@
...
+    "\n",
+    "<div class=\"alert alert-info\">\n",
+    "\n",
+    "As an example, let's say a user splits their area of interest into `10.000` smaller bounding boxes. Then they parallelize data download from Sentinel Hub over these bounding boxes using `10` different Python processes. It depends on the type of parallelization, but without implementing a session sharing mechanism such a procedure can create `10.000` different authentication sessions. This can cause the user to be temporary restricted from using Sentinel Hub services. In such cases please contact [support](https://docs.sentinel-hub.com/api/latest/support/).\n",
+    "</div>\n",
+    "\n",
+    "\n",
+    "This tutorial will explain the basic mechanism of working with sessions in `sentinelhub-py` and how to implement a session transfer for different types of parallelization. \n",
+    "\n",
+    "For the purposes of the tutorial, let's first configure a logging handler that will tell us exactly when a new session has been created:"
```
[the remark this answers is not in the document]
AleksMat: I just bolded it. Additionally capitalizing it might leave the impression of "screaming level extreme" importance which is not. :)

examples/session_sharing.ipynb
AleksMat: Fixed.

PR #281, opened 2022-05-27
Improve the way we build docs

zigaLuksic:
Does the following:

1. Fix `.readthedocs.yaml`
2. Enable autogenerated .rst files
    - to work on `ReadTheDocs` it is triggered in `conf.py`
    - they use custom template files, otherwise the toctrees become too large and unreadable
    - enabled option to transfer type annotations to parameter docstrings, to produce more nicely structured signatures (same as before)
3. Removed files that are no longer needed
4. Removed version bounds in `requirements-docs.txt` as they are no longer needed


docs/source/conf.py
```diff
@@ -253,3 +257,22 @@ def process_readme():
 
 
 process_readme()
+
+# Auto-generate documentation pages
+current_dir = os.path.abspath(os.path.dirname(__file__))
+module = os.path.join(current_dir, "..", "..", "sentinelhub")
+
+APIDOC_EXCLUDE = [os.path.join(module, "commands.py")]
```
AleksMat, line 265: I would also put `sentinelhub.aws.commands` module to this list.

requirements-docs.txt
```diff
@@ -1,9 +1,10 @@
...
+docutils
 sphinx
 sphinx_rtd_theme
 nbsphinx
 matplotlib
 ipython
 m2r2
-jinja2==3.0.3
+jinja2
+pytest
```
AleksMat, line 10: Since `pytest` is on `autodoc_mock_imports` list it probably doesn't have to be here.

docs/source/conf.py
```diff
@@ -253,3 +257,22 @@ def process_readme():
...
+APIDOC_OPTIONS = ["--module-first", "--separate", "--no-toc", "--templatedir", os.path.join(current_dir, "_templates")]
+
+
+def run_apidoc(_):
+    from sphinx.ext.apidoc import main
+
+    sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
+
+    main(["-e", "-o", current_dir, module, *APIDOC_EXCLUDE, *APIDOC_OPTIONS])
```
AleksMat, line 274: Is there a way that auto-generated files would be deleted at the end of this procedure? The problem is that when running docs build locally it leaves a lot of these files that we would probably accidentally be committing to git. 

Another option would be to put `sentinelhub.*.rst` files to `.gitignore`. If at any point we will want to commit any of them we'll commit with `-f`. Perhaps we could even put all of them into a subfolder of `source` folder so that they don't clutter the main docs folder?

docs/source/conf.py
```diff
@@ -98,7 +102,7 @@
 todo_include_todos = True
 
 # Mock imports that won't and don't have to be installed in ReadTheDocs environment
-autodoc_mock_imports = ["boto3", "botocore"]
+autodoc_mock_imports = ["boto3", "botocore", "pyproj", "pytest"]
```
AleksMat, line 105: Do we still have to include `pyproj` here or has that issue been resolved?

requirements-docs.txt
zigaLuksic: i actually believe that `pytest` needs to be included in the regular requirements, since we use it in `testing_utils.py`

docs/source/conf.py
zigaLuksic: I would prefer to do it via .gitignore, otherwise we again need to handcraft a list of what to remove and what not to remove.
moving them around is difficult since the links between them are auto-generated

docs/source/_templates/package.rst_t
```diff
@@ -0,0 +1,52 @@
+{%- macro automodule(modname, options) -%}
```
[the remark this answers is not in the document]
zigaLuksic: I would keep template edits to a minimum, since it makes it easier to compare to the originals.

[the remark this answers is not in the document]
zigaLuksic: also judging from the templates, it looks to me like whitespace persists, so it might change the output .rst files

requirements-docs.txt
AleksMat: No, we can always say that `sentinelhub.testing_utils` is a part of `DEV` package extension. I would definitely not add it to the main requirements which are already bloated.

docs/source/conf.py
AleksMat: Ok, I agree.

It might work if you put all of them and `sentinelhub.rst` into a subfolder. The only reference you would have to modify then is the reference from `index.rst` to `sentinelhub.rst`.

requirements-docs.txt
zigaLuksic: works without it, removed

docs/source/conf.py
zigaLuksic: both of these can be removed

docs/source/conf.py
zigaLuksic: done

docs/source/_templates/package.rst_t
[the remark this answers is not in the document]
AleksMat: Ok, let's leave it for now.

docs/source/conf.py
AleksMat: Looks great. :bowing_man:

PR #283, opened 2022-06-01
Fixes for CI and pylint

AleksMat:
Changes:

- It turns out our GitHub action was missing branch checkout in case of scheduled runs for some jobs. I fixed this by parametrizing the branch that gets checked out and only kept 1 checkout step in each job.

- Pylint released a new version with a couple of additional checks. One of them was complaining that `pylintrc` contains come configurations that are not supported (anymore?). So I deleted the massive `pylintrc` file and moved only the non-default parameters into `pyproject.toml`.


zigaLuksic (approved):
Love the removal of pylintrc. I'm just sad that flake8 cannot be moved to pyproject.toml yet :(

PR #284, opened 2022-06-02
Improve handling of unclosed shared memory

AleksMat:
This PR adds 2 improvements:

- `SessionSharingThread` itself is now able to handle unclosed shared memory and warn users about it.
- `SessionSharing` object that is a more user-friendly interface around `SessionSharingThread`.

### DIFF
--- a/tests/download/test_session.py
+++ b/tests/download/test_session.py
@@ -121,3 +121,17 @@ def test_session_sharing_multiprocess(fake_token, fake_config, memory_name):
         assert all(collected_session.token == fake_token for collected_session in collected_sessions)
     finally:
         thread.join()
+
+
+def test_handling_of_unclosed_memory(fake_token, fake_config):
+    session = SentinelHubSession(config=fake_config, refresh_before_expiry=0, _token=fake_token)
+
+    thread1 = SessionSharingThread(session)
+    thread1.start()
+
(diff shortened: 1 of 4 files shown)


tests/download/test_session.py
```diff
@@ -121,3 +121,35 @@ def test_session_sharing_multiprocess(fake_token, fake_config, memory_name):
...
+    manager = SessionSharing(session, name="thread name", **kwargs)
+
+    with pytest.raises(FileNotFoundError):
+        collect_shared_session(**kwargs)
+
+    with manager:
+        collected_session = collect_shared_session(**kwargs)
+        assert collected_session.token == fake_token
+
+    with pytest.raises(FileNotFoundError):
```
zigaLuksic, line 140: do we want this function to raise a more informative error?
not sure, it feels like we're improving this way beyond what is needed/expected, but also it feels wrong to leave users confused :see_no_evil:

zigaLuksic (approved):
Since we provide a manager and use that as the go-to in the tutorial, I don't mind the 'override if present' behaviour. Great job.

tests/download/test_session.py
AleksMat: Good point, I added that the same `FileNotFoundError` type gets raised but with a better error message.

PR #286, opened 2022-06-10
Compliance hook for Sentinel Hub authentication

AleksMat:
Package `requests-oauthlib` doesn't check for status errors of a response obtained from Sentinel Hub authentication service. Because of that it can raise a misleading `MissingTokenError`.

This PR solves the issue by adding a compliance hook into the process which checks for a response status and makes sure the correct error is raised. It also adds units tests for it.

### DIFF
--- a/sentinelhub/download/session.py
+++ b/sentinelhub/download/session.py
@@ -149,7 +149,7 @@ class SentinelHubSession:
         without an error message. In such cases `requests_oauthlib` would raise a completely wrong error message. This
         hook makes sure that a correct error message is raised.
 
-        It is important that in case of 5xx errors always an error is raised so that authentication can be retried.
+        It is important that in case of 5xx errors an error is always raised so that authentication can be retried.
         But in case of 4xx errors where response contains an error message this method intentionally doesn't raise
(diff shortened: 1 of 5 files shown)


sentinelhub/download/session.py
```diff
@@ -131,10 +134,38 @@ def _fetch_token(self, request: DownloadRequest) -> JsonDict:
...
+    @staticmethod
+    def _compliance_hook(response: Response) -> Response:
+        """Checks if a response from Sentinel Hub Authentication service has an error status code but no error message.
+
+        By default, `requests_oauthlib` ignores status of a response and only looks at an error message in a
+        response body. However, Sentinel Hub service can return a response with an error status code and
+        without an error message. In such cases `requests_oauthlib` would raise a completely wrong error message. This
+        hook makes sure that a correct error message is raised.
+
+        It is important that in case of 5xx errors always an error is raised so that authentication can be retried.
```
zigaLuksic, line 152:
```suggestion
        It is important that in case of 5xx errors an error is always raised so that authentication can be retried.
```

zigaLuksic (approved):

PR #288, opened 2022-06-13
Tests for download handlers and improved error message parsing

AleksMat:
Increasing test coverage and improving the way error messages are being parsed.

### DIFF
--- a/sentinelhub/decoding.py
+++ b/sentinelhub/decoding.py
@@ -164,7 +164,7 @@ def decode_sentinelhub_err_msg(response: Response) -> str:
         for elem in decode_data(response.content, MimeType.XML):
             if "ServiceException" in elem.tag or "Message" in elem.tag or elem.tag == "body":
                 for text in elem.itertext():
-                    stripped_text = text.strip("\n\t ")
+                    stripped_text = text.strip()
                     if stripped_text:
                         server_message.append(stripped_text)
         return " ".join(server_message)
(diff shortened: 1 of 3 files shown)


sentinelhub/decoding.py
```diff
@@ -144,12 +145,29 @@ def decode_sentinelhub_err_msg(response: Response) -> str:
...
+
     try:
         server_message = []
         for elem in decode_data(response.content, MimeType.XML):
-            if "ServiceException" in elem.tag or "Message" in elem.tag:
-                server_message.append(elem.text.strip("\n\t "))
-        return "".join(server_message)
+            if "ServiceException" in elem.tag or "Message" in elem.tag or elem.tag == "body":
+                for text in elem.itertext():
+                    stripped_text = text.strip("\n\t ")
```
zigaLuksic, line 167: `strip` already removes all whitespace characters. Is there a reason these are specified by hand?

tests/download/test_handlers.py
```diff
@@ -0,0 +1,123 @@
...
+    dummy_client = DummyClient()
+
+    @retry_temporary_errors
+    def fail_process(client: DummyClient, _: DownloadRequest) -> None:
+        client.increase_count()
+        raise exception
+
+    expected_exception_class = DownloadFailedException if is_expected_handling else exception.__class__
+    with pytest.raises(expected_exception_class):
+        fail_process(dummy_client, DownloadRequest())
```
zigaLuksic, line 57: do we need to use `DownloadRequest` here (in this test module)? seems irrelevant to tests

sentinelhub/decoding.py
zigaLuksic: i had to doublecheck as well :smile:

PR #291, opened 2022-06-17
Extend CI to also lint example notebooks

zigaLuksic:
By using `nbqa` we can also lint the example notebooks. Pylint is problematic in that regard, so I added `flake8` configuration (two linters are better than none).

Since we now use many linters I made one runner do all the 'no-install-required' linting.

We could also run `mypy` on the notebooks, but that would require a fair amount of notebook refactoring.


requirements-dev.txt
```diff
@@ -11,6 +11,7 @@ click>=8.0.0
 rasterio;python_version<"3.10"
 ray[default]
 mypy
+nbqa
```
AleksMat, line 14: Let's also put `flake8` on the list.

.github/workflows/ci_action.yml
```diff
@@ -11,31 +11,30 @@ env:
...
-        with:
-          ref: ${{ env.CHECKOUT_BRANCH }}
+      - name: Check code compliance with black
+        run: black . --check --diff
 
       - name: Check code compliance with isort
         run: |
-          pip install isort
           isort . --check --diff
+          nbqa isort examples --check --diff
```
AleksMat, line 31: You could actually do `nbqa isort . --check --diff` because this will recursively find and check all notebooks in all subfolders. Just in case we ever add any other folder with notebooks.

.github/workflows/ci_action.yml
```diff
@@ -11,31 +11,30 @@ env:
...
+
+      - name: Check code compliance with flake8
+        run: |
+          pip install isort
+          flake8 .
+          nbqa flake8 examples --ignore=E402
```
AleksMat, line 37: - `pip install isort` should be removed,
- I would also here change `examples` to `.` so that we have `nbqa flake8 . --ignore=E402`

.github/workflows/ci_action.yml
MarcoGorelli: hey thanks for trying out nbQA - I'd suggest using `--nbqa-diff` instead of isort's `--diff`, it'll just look a bit better with notebooks

zigaLuksic: @MarcoGorelli  Thanks for the tip!

PR #295, opened 2022-07-07
Switch back to pillow for jp2

zigaLuksic:


AleksMat (changes requested):
Seems ok, but also please remove `rasterio` from:

- `requirements-dev.txt`,
- `test_data.py`,
- `test_io_utils.py`

PR #301, opened 2022-08-04
Add Batch Statistical API

zigaLuksic:
Adds the Batch Statistical API to sh-py.

I did some light generalization of batch process & batch statistical, but as discussed privately, generalizing further would hurt docstrings.

Generalizing 'monitor_batch_job' and writing an example notebook would be done in future MR


sentinelhub/api/batch_statistical.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class SentinelHubBatchStatistical(BaseBatchClient["BatchStatisticalRequest"]):
+    """An interface class for Sentinel Hub Batch Statistical API
+
+    Check `Batch Statistical API <https://docs.sentinel-hub.com/api/latest/reference/#tag/batch_statistical>`__ for more
+    information.
+    """
+
+    # pylint: disable=too-many-public-methods
```
AleksMat, line 30: I think you don't need this statement here because in this case there are only 6 public methods.

sentinelhub/api/batch_statistical.py
```diff
@@ -0,0 +1,148 @@
...
+            "calculations": calculations,
+            "output": output,
+            **kwargs,
+        }
+        payload = remove_undefined(payload)
+
+        url = self.service_url
+        request_info = self.client.get_json(url, post_values=payload, use_session=True)
+
+        return BatchStatisticalRequest.from_dict(request_info)  # type: ignore[attr-defined]
```
AleksMat, line 63: Do you think we could add just a signature for methods `from_dict` and `to_dict` to all classes that are wrapped with `@dataclass_json`? This way the typing and autocomplete would start working.

sentinelhub/api/batch_base.py
```diff
@@ -0,0 +1,98 @@
+"""
```
AleksMat, line 1: I'm wondering if module name `base_batch.py` would be better. This way it would be more in-line with `base.py` and `base_request.py`.

sentinelhub/api/batch_statistical.py
```diff
@@ -0,0 +1,148 @@
...
+    def _get_service_url(base_url: str) -> str:
+        """Provides URL to Batch Statistical API"""
+        return f"{base_url}/api/v1/statistics/batch"
+
+    def create(
+        self,
+        input_data: JsonDict,
+        input_features: JsonDict,
+        aggregation: JsonDict,
+        calculations: JsonDict,
```
AleksMat, line 41: I'm starting to wonder if perhaps it would be better to replace `input_data`, `aggregation`, and `calculations` with `statistical_request`. This would be more in-line with `SentinelHubBatch`. :thinking:

sentinelhub/api/batch_statistical.py
```diff
@@ -0,0 +1,148 @@
...
+
+        `Batch Statistical API reference
+        <https://docs.sentinel-hub.com/api/latest/reference/#operation/createNewBatchStatisticsRequest>`__
+        """
+
+        payload = {
+            "input": {"features": input_features, "data": input_data},
+            "aggregation": aggregation,
+            "calculations": calculations,
+            "output": output,
```
AleksMat, line 55: I see that this schema for an s3 bucket appears both in `input_features` and `output`. Perhaps we could have a helper method for building this payload e.g. `SentinelHubBatchStatistical.s3` or something more verbose. We have a similar helper methods also in the other batch class.

sentinelhub/api/batch_statistical.py
```diff
@@ -0,0 +1,148 @@
...
+from ..type_utils import Json, JsonDict
+from .batch_base import BaseBatchClient, BaseBatchRequest, BatchRequestStatus, BatchUserAction
+from .utils import datetime_config, enum_config, remove_undefined
+
+LOGGER = logging.getLogger(__name__)
+
+BatchStatisticalRequestType = Union[str, dict, "BatchStatisticalRequest"]
+
+
+class SentinelHubBatchStatistical(BaseBatchClient["BatchStatisticalRequest"]):
```
AleksMat, line 23: Because the service is called Statistical Batch, should we change the name of this class to `SentinelHubStatisticalBatch`?

sentinelhub/api/batch_statistical.py
```diff
@@ -0,0 +1,148 @@
...
+    request: dict
+    status: BatchRequestStatus = field(metadata=enum_config(BatchRequestStatus))
+    user_id: Optional[str] = None
+    created: Optional[dt.datetime] = field(metadata=datetime_config, default=None)
+    cost_pu: Optional[float] = None
+    user_action: Optional[BatchUserAction] = field(metadata=enum_config(BatchUserAction), default=None)
+    user_action_updated: Optional[str] = field(metadata=datetime_config, default=None)
+    error: Optional[str] = None
+
+    _REPR_PARAM_NAMES = (
```
AleksMat, line 141: Also maybe `completion_percentage` could be put to this list?

sentinelhub/api/batch_statistical.py
```diff
@@ -0,0 +1,148 @@
...
+
+    # pylint: disable=too-many-public-methods
+    @staticmethod
+    def _get_service_url(base_url: str) -> str:
+        """Provides URL to Batch Statistical API"""
+        return f"{base_url}/api/v1/statistics/batch"
+
+    def create(
+        self,
+        input_data: JsonDict,
```
AleksMat, line 38: `input_data` should actually have a type `List[JsonDict]`. In the process and statistical API interfaces we even use type `List[Union[JsonDict, InputDataDict]]` for this so maybe this an even better type.

sentinelhub/api/batch_base.py
```diff
@@ -0,0 +1,98 @@
...
+        if isinstance(data, BaseBatchRequest):
+            return data.request_id
+        if isinstance(data, dict):
+            return data["id"]
+        if isinstance(data, str):
+            return data
+        raise ValueError(f"Expected a BatchRequest, dictionary or a string, got {data}.")
+
+
+class BaseBatchRequest:
```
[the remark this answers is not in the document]
zigaLuksic: because dataclasses + inheritance + default values dont mix well. If the parent dataclass has an attribute with a default value then the child classes can only extend with default-valued attributes.

I could wrap this in a dataclass and just leave `error` without a default and then override the field with a new field + default in the child classes.

sentinelhub/api/batch_statistical.py
zigaLuksic: this was part of the discussion we had previously. I was thinking that we could perhaps add a builder method that would work with `statistical_request` +  S3 params?

sentinelhub/api/batch_statistical.py
zigaLuksic: The official docs keep switching between [Statistical Batch](https://docs.sentinel-hub.com/api/latest/reference/#tag/batch_statistical) and [Batch Statistical](https://docs.sentinel-hub.com/api/latest/api/batch-statistical/)

I went with `BatchStatistical` since it occured more often

sentinelhub/api/batch_base.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _call_job(self, batch_request: RequestSpec, endpoint_name: str) -> Json:
+        """Makes a POST request to the service that triggers a processing job"""
+        request_id = self._parse_request_id(batch_request)
+        job_url = f"{self._get_job_endpoint_url(request_id)}/{endpoint_name}"
+        return self.client.get_json(url=job_url, request_type=RequestType.POST, use_session=True)
+
+    def _get_job_endpoint_url(self, request_id: Optional[str] = None) -> str:
+        """Creates a URL for the batch statistical endpoint"""
+        url = f"{self.service_url}"
```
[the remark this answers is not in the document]
zigaLuksic: done

sentinelhub/api/batch_statistical.py
zigaLuksic: done

sentinelhub/api/batch_statistical.py
zigaLuksic: done

sentinelhub/api/batch_statistical.py
zigaLuksic: done

sentinelhub/api/batch.py
```diff
@@ -456,16 +462,7 @@ def _get_collections_url(self, collection_id: Optional[str] = None) -> str:
...
-    def _parse_request_id(data: BatchRequestType) -> str:
-        """Parses batch request id from multiple possible inputs"""
-        if isinstance(data, BatchRequest):
-            return data.request_id
-        if isinstance(data, dict):
-            return data["id"]
-        if isinstance(data, str):
-            return data
-        raise ValueError(f"Expected a BatchRequest, dictionary or a string, got {data}.")
+    _get_job_endpoint_url = _get_process_url
```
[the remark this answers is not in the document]
zigaLuksic: I was thinking of the first option, but then didn't decide for it since in batch process there is a distinction between grids, process and collections. However on second look it seems like the renaming wouldn't hurt readability all that much

sentinelhub/api/batch_base.py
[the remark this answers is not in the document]
AleksMat: Ah, you are right. This is an annoying thing with dataclasses.

But even now you don't assign `error` attribute a default value. So in the current implementation it might work?

sentinelhub/api/batch.py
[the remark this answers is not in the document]
zigaLuksic: went with second option

sentinelhub/api/batch_statistical.py
AleksMat: I would still leave `input_features` because `s3` might not be the only option ever supported. But maybe `(statistical_request, input_features, output, **kwargs)` could work :thinking:

sentinelhub/api/batch_statistical.py
AleksMat: Ah, I see. This is something we could mention to the core team. :see_no_evil: 

IMO, it is nicer if adjective comes before noun in the name but we should follow what the core team decides.

zigaLuksic: Well it think that is the problem, because both of them could be used as the adjective
Batch Statistical (as in statistical done in a batch approach)
Statistical Batch (batch download of statistical nature)

sentinelhub/api/batch_base.py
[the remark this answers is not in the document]
zigaLuksic: sadly this couldn't be done (easily). I tried many options but dataclass_json seems very inflexible when it comes to inheritance

sentinelhub/api/batch_statistical.py
zigaLuksic: done

sentinelhub/api/batch_statistical.py
```diff
@@ -0,0 +1,149 @@
...
+    request: dict
+    status: BatchRequestStatus = field(metadata=enum_config(BatchRequestStatus))
+    user_id: Optional[str] = None
+    created: Optional[dt.datetime] = field(metadata=datetime_config, default=None)
+    cost_pu: Optional[float] = None
+    user_action: Optional[BatchUserAction] = field(metadata=enum_config(BatchUserAction), default=None)
+    user_action_updated: Optional[str] = field(metadata=datetime_config, default=None)
+    error: Optional[str] = None
+
+    _REPR_PARAM_NAMES = (
```
AleksMat, line 141: I see you are using a tuple here and it is better than a list in `BatchRequest` class. Would you mind changing to tuple there as well?

zigaLuksic: done

sentinelhub/api/batch_statistical.py
AleksMat: After checking with the core team we are keeping the order Batch Statistical.

sentinelhub/api/batch_statistical.py
zigaLuksic: done

sentinelhub/api/batch_base.py
AleksMat: Moving things into a subfolder did the job :+1:

sentinelhub/api/batch/statistical.py
```diff
@@ -0,0 +1,215 @@
...
+class S3AccessSpecification(TypedDict):
+    """Specification of a S3 path."""
+
+    url: str
+    accessKey: str
+    secretAccessKey: str
+    region: NotRequired[str]
+
+
+class S3Specification(TypedDict):
```
AleksMat, line 40: Shouldn't this class be called maybe `StorageSpecification` or `AccessSpecification` and the one above `S3Specification`? This is because here `s3` is only one of the parameters (atm the only one).

sentinelhub/api/batch/statistical.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def create(
+        self,
+        input_features: S3Specification,
+        input_data: List[Union[JsonDict, InputDataDict]],
+        aggregation: JsonDict,
+        calculations: JsonDict,
+        output: S3Specification,
+        **kwargs: Any,
+    ) -> "BatchStatisticalRequest":
```
AleksMat, line 66: If you used

```Python
from __future__ import annotations
```
the return type here could be `BatchStatisticalRequest` instead of `"BatchStatisticalRequest"`, right? Although I'm not sure if that would make it in any way better than now.

(Same comment applies fro the method below.)

sentinelhub/api/batch/statistical.py
```diff
@@ -0,0 +1,215 @@
...
+            **kwargs,
+        }
+        payload = remove_undefined(payload)
+
+        url = self.service_url
+        request_info = self.client.get_json_dict(url, post_values=payload, use_session=True)
+
+        return BatchStatisticalRequest.from_dict(request_info)
+
+    def create_from_statistical_request(
```
AleksMat, line 87: Maybe the name could be shortened to `create_from_request`? Because it is should be rather clear it is a statistical request.

Although on the other hand we also have `BatchStatisticalRequest` class which could be confused with `SentinelHubStatistical`. This naming is a nightmare. :see_no_evil:

AleksMat (approved):
I wrote only a few minor comments, otherwise the PR is good to merge. :+1:

sentinelhub/api/batch/statistical.py
zigaLuksic: done

sentinelhub/api/batch/statistical.py
```diff
@@ -0,0 +1,215 @@
...
+    region: NotRequired[str]
+
+
+class AccessSpecification(TypedDict):
+    """Specification of a S3 input or output."""
+
+    s3: S3Specification
+
+
+class SentinelHubBatchStatistical(BaseBatchClient["BatchStatisticalRequest"]):
```
chorng, line 46: Could we add a job monitor method similar to batch to the class? I have a code snippet in [cell 20 of the notebook](https://git.sinergise.com/sh-vas/technical-support/clients/-/blob/master/SuperAlberi/how_to_obtain_statistics_with_batch_statistical_api.ipynb), which update the progress bar by `completionPercentage`. An alternative way will be checking the files in the bucket and update by `files in the bucket / length of .gpkg file`

chorng: Sorry... click the wrong option. Just wanted to add a comment... 🙈

zigaLuksic: This will be added in a separate MR (so that we keep them a bit smaller and incremental)

chorng: Thanks! May I ask if we plan to add other methods that extract statistics to a data frame from s3?

PR #302, opened 2022-08-06
DownloadResponse and more download-related improvements

AleksMat:
This PR grew a bit too large because it required quite a significant change in the download procedure. But I'm satisfied with how it turned out and with additional improvements I've made.

I had to make a few code-breaking changes but only to less used functionalities:

- Specifying `decode_data=False` in `DownloadClient.download` or `DataRequest.get_data` methods previously returned a response in `bytes` but now it returns a `DownloadResponse` object which has an attribute `content` which is the response in `bytes`.
- A structure of `request.json` cache file has changed and now contains not only request parameters but also response metadata parameters. But I made sure that "cache matching" procedure `DownloadClient` is backwards compatible.
- I converted `DownloadRequest` into a dataclass to be in-line with the new `DataResponse` class. Because of that we couldn't have `**properties` kwargs in `__init__` anymore and I turned it into `extra_params` parameter. I think this is a safer implementation.

I had some further ideas which would be even more code-breaking and I didn't implement:

- We could change the name of cache file `request.json` into either `info.json` or `metadata.json` because the new version contains info about both the request and the response.
- I've already renamed `request.py` module into `models.py` module. This is inspired by the `requests.models` module which contains `requests.Request` and `requests.Response` objects. We could even rename `DownloadRequest` and `DownloadResponse` into `RequestModel` and `ResponseModel`. Advantages:
  * This way there would be less confusion between `DownloadRequest` and `DataRequest` naming.
  * The word "download" is also not completely appropriate as these classes can be used for any kind of HTTP interaction, including upload.
- Class `DownloadRequest` could inherit from `requests.Request` but that would require that we rename some of its parameter, e.g. `post_values`, which we use very often. I'm also not 100% sure how safe such inheritance would be. 

Any of these ideas could potentially be done in a future PR.


sentinelhub/download/client.py
```diff
@@ -89,16 +88,18 @@ def download(
         :param download_requests: A list of requests or a single request to be executed.
```
zigaLuksic, line 88: Are `DownloadRequest` and `DownloadResponse` ever specialized in when subclassing download client? If so, I believe it would be nice if we turn it into a generic. This can (probably should) be done in a separate MR

sentinelhub/download/models.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def get_hashed_name(self) -> str:
+        """It takes request url and payload and calculates a unique hashed string from them.
+
+        :return: A hashed string
+        """
+        params = self.get_request_params(include_metadata=False)
+        hashable = json.dumps(params)
+
+        return hashlib.md5(hashable.encode("utf-8")).hexdigest()
```
zigaLuksic, line 98: This change will change all the hashes, right? Perhaps to keep in mind when writing a changelog

sentinelhub/download/models.py
```diff
@@ -0,0 +1,257 @@
...
+        return cls(
+            request=request,
+            content=response.content,
+            headers=dict(response.headers),
+            status_code=response.status_code,
+            elapsed=response.elapsed.total_seconds(),
+        )
+
+    @classmethod
+    def from_local(cls, request: DownloadRequest) -> DownloadResponse:
```
zigaLuksic, line 186: do we care about the case where the response is saved locally but the `request` object is no longer in python memory? So loading directly from path?

sentinelhub/download/client.py
AleksMat: Specialized in terms of what type of data they will download and decode? There are actually a lot of options, usually it depends on `data_type` parameter although, if not provided, the type can also be decided from response headers.

So it would be really tricky to turn this into generics.

sentinelhub/download/models.py
AleksMat: Nope, the hashes will stay exactly the same.

sentinelhub/download/client.py
zigaLuksic: specialized more as 'subclassed'. So if we ever have `BolhaDownloadClient` with `BolhaDownloadResponse` instead of regular just the `DownloadResponse`.

sentinelhub/download/models.py
zigaLuksic: ah, because `extra_params` aren't included in the `get_request_params`

sentinelhub/download/models.py
AleksMat: The problem is that any instance of `DownloadResponse` object should contain an instance of `DownloadRequest` object as its attribute. This is due to various practical reasons. So loading data directly from a local path would mean we would have to reconstruct the instance of `DownloadRequest` object from the cached content as well. It is possible to implement but atm, I don't see a need to do it.

sentinelhub/download/client.py
AleksMat: Sure.

PR #303, opened 2022-08-07
A few docs-related updates

AleksMat:
Changes:

- The main change is how auto-generated reference files are handled. The problem with the previous version is that `reference` folder contained both committed and auto-generated files. This meant we couldn't remove auto-generated files in-between docs build runs. Hence, outdated auto-generate files would remain in `reference` folder and this would cause problems. (Note this is only an issue when someone is building docs locally.)

- A new version of Sphinx was complaining about `language = None` settings so I changed it according to its recommendation.

- I noticed we can add `project_urls` into setup and [here](https://test.pypi.org/project/sentinelhub/3.6.4.1/) you can see how this change looks once released to Test PyPI.

If this gets approved I can make same changes for `eo-learn` and `eo-grow`.


docs/source/conf.py
```diff
@@ -261,19 +261,23 @@ def process_readme():
 
 # Auto-generate documentation pages
 current_dir = os.path.abspath(os.path.dirname(__file__))
-target_dir = os.path.join(current_dir, "reference")
+reference_dir = os.path.join(current_dir, "reference")
+predefined_reference_dir = os.path.join(current_dir, "_reference")
```
zigaLuksic, line 265: Could we have a better name than `_reference`? Perhaps `custom_reference` or something? It would make more sense that the auto-generated reference folder would be `_reference` IMO

zigaLuksic (commented):
If I understood correctly 'handcrafted' pages are n in `_reference`

docs/source/conf.py
AleksMat: Ok, I renamed to `custom_reference`.

PR #307, opened 2022-08-07
Added apply method to geometry objects

AleksMat:


sentinelhub/geometry.py
```diff
@@ -1,9 +1,11 @@
 """
 Module implementing geometry classes
 """
+from __future__ import annotations
```
zigaLuksic, line 4: this is just to avoid `"BBox"` and the likes? I don't really mind

zigaLuksic (approved):

sentinelhub/geometry.py
AleksMat: Yes, it seems nicer to me, but I don't know if it brings any benefit at all.

zigaLuksic: aside from aesthetic ones I don't think so. At least not in this case

PR #309, opened 2022-08-08
Adjust monitoring for batch statistical

zigaLuksic:
This approach follows the current sh-py ideology "we take care of everything".
Sadly this results in multiple heavy duty overloads.

The current MR is more of a showcase of how it can be done, but we should consider switching to two separate functions. There will be more copied code but less type-system magic

### DIFF
--- a/sentinelhub/api/batch/utils.py
+++ b/sentinelhub/api/batch/utils.py
@@ -2,93 +2,55 @@
 Module implementing utilities for working with batch jobs.
 """
 import logging
-import sys
 import time
 import warnings
 from collections import defaultdict
-from typing import DefaultDict, List, Optional, TypeVar, Union, overload
+from typing import DefaultDict, List, Optional, Union
 
 from tqdm.auto import tqdm
 
 from ...config import SHConfig
+from ...exceptions import SHDeprecationWarning
+from ...type_utils import JsonDict
 from .base import BatchRequestStatus
 from .process import BatchRequest, BatchTileStatus, SentinelHubBatch
 from .statistical import BatchStatisticalRequest, SentinelHubBatchStatistical
 
-if sys.version_info < (3, 8):
-    from typing_extensions import Literal
-else:
-    from typing import Literal  # pylint: disable=ungrouped-imports
-
 LOGGER = logging.getLogger(__name__)
 
-BatchRequestSpec = Union[str, dict, BatchRequest, BatchStatisticalRequest]
-BatchRequestType = TypeVar("BatchRequestType", BatchStatisticalRequest, BatchRequest)  # pylint: disable=invalid-name
-BatchKind = Literal["process", "statistical"]
+BatchProcessRequestSpec = Union[str, dict, BatchRequest]
+BatchStatisticalRequestSpec = Union[str, dict, BatchStatisticalRequest]
 
 
 _MIN_SLEEP_TIME = 60
 _DEFAULT_SLEEP_TIME = 120
+_MIN_STAT_SLEEP_TIME = 10
+_DEFAULT_STAT_SLEEP_TIME = 15
 _MIN_ANALYSIS_SLEEP_TIME = 5
 _DEFAULT_ANALYSIS_SLEEP_TIME = 10
 
 
-@overload
 def monitor_batch_job(
-    batch_request: Union[str, dict],
-    config: Optional[SHConfig],
-    sleep_time: int = _DEFAULT_SLEEP_TIME,
-    analysis_sleep_time: int = _DEFAULT_ANALYSIS_SLEEP_TIME,
-    *,
-    batch_kind: Literal["process"],
-) -> DefaultDict[BatchTileStatus, List[dict]]:
-    pass
-
-
-@overload
-def monitor_batch_job(
-    batch_request: BatchRequest,
-    config: Optional[SHConfig],
-    sleep_time: int = _DEFAULT_SLEEP_TIME,
-    analysis_sleep_time: int = _DEFAULT_ANALYSIS_SLEEP_TIME,
-    *,
-    batch_kind: BatchKind,
-) -> DefaultDict[BatchTileStatus, List[dict]]:
-    pass
-
-
-@overload
-def monitor_batch_job(
-    batch_request: Union[str, dict],
-    config: Optional[SHConfig],
-    sleep_time: int = _DEFAULT_SLEEP_TIME,
-    analysis_sleep_time: int = _DEFAULT_ANALYSIS_SLEEP_TIME,
-    *,
-    batch_kind: Literal["statistical"],
-) -> None:
-    pass
-
-
-@overload
-def monitor_batch_job(
-    batch_request: BatchStatisticalRequest,
-    config: Optional[SHConfig],
-    sleep_time: int = _DEFAULT_SLEEP_TIME,
-    analysis_sleep_time: int = _DEFAULT_ANALYSIS_SLEEP_TIME,
-    *,
-    batch_kind: BatchKind,
-) -> None:
-    pass
-
-
-def monitor_batch_job(
-    batch_request: BatchRequestSpec,
+    batch_request: BatchProcessRequestSpec,
     config: Optional[SHConfig] = None,
     sleep_time: int = _DEFAULT_SLEEP_TIME,
     analysis_sleep_time: int = _DEFAULT_ANALYSIS_SLEEP_TIME,
-    *,
-    batch_kind: BatchKind = "process",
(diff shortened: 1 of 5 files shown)


sentinelhub/api/batch/utils.py
```diff
@@ -2,31 +2,93 @@
...
+    config: Optional[SHConfig],
+    sleep_time: int = _DEFAULT_SLEEP_TIME,
+    analysis_sleep_time: int = _DEFAULT_ANALYSIS_SLEEP_TIME,
+    *,
+    batch_kind: BatchKind,
+) -> None:
+    pass
+
+
+def monitor_batch_job(
```
AleksMat, line 84: Because of the completely different return type I would suggest that `monitor_batch_job` is not overloaded and we instead have a separate `monitor_batch_statistical_job` function. Also the default sleep time could be much shorter for statistical batch because we don't have to make a ton of request every time.

On the other hand, `monitor_batch_analysis` seems similar enough that we can keep the same implementation.

sentinelhub/api/batch/utils.py
```diff
@@ -50,11 +112,77 @@ def monitor_batch_job(
...
+) -> BatchStatisticalRequest:
+    pass
+
+
+def monitor_batch_analysis(
+    batch_request: BatchRequestSpec,
+    config: Optional[SHConfig] = None,
+    sleep_time: int = _DEFAULT_ANALYSIS_SLEEP_TIME,
+    *,
+    batch_kind: BatchKind = "process",
```
AleksMat, line 153: I would actually put a default `batch_kind: BatchKind = "auto"`. In this case, the kind would be determined from `batch_request` (if class is given) or an error would be raised.

sentinelhub/api/batch/utils.py
```diff
@@ -50,11 +112,77 @@ def monitor_batch_job(
...
+            f"To avoid making too many service requests please set analysis sleep time >={_MIN_ANALYSIS_SLEEP_TIME}"
+        )
+    batch_kind = _auto_adjust_kind(batch_request, batch_kind)
+
+    batch_client = _get_batch_client(batch_kind, config)
+    batch_request = batch_client.get_request(batch_request)  # type: ignore[arg-type]
+    while batch_request.status in [BatchRequestStatus.CREATED, BatchRequestStatus.ANALYSING]:
+        LOGGER.info("Batch job has a status %s, sleeping for %d seconds", batch_request.status.value, sleep_time)
+        time.sleep(sleep_time)
+        batch_request = batch_client.get_request(batch_request)  # type: ignore[arg-type]
```
AleksMat, line 177: In case of a statistical batch this could actually make a call to status endpoint instead of the main endpoint. This way the size of the payload would be much smaller.

However, this would mean that we might also want that this function returns only the status payload. Because of that we might even decide to have separate functions for statistical batch even for this part.

sentinelhub/api/batch/utils.py
```diff
@@ -86,39 +214,28 @@ def monitor_batch_job(
...
-    :param sleep_time: Number of seconds between consecutive status updates during analysis phase.
-    :return: Batch request info
-    """
-    if sleep_time < _MIN_ANALYSIS_SLEEP_TIME:
-        raise ValueError(
-            f"To avoid making too many service requests please set analysis sleep time >={_MIN_ANALYSIS_SLEEP_TIME}"
-        )
+    def current_completion_percentage() -> float:
+        """Fetches the current completion percentage"""
+        return batch_client.get_status(batch_request)["completionPercentage"]
```
AleksMat, line 224: I find it unnecessary to have a separate function for `1` line of code.

sentinelhub/api/batch/utils.py
```diff
@@ -86,39 +214,28 @@ def monitor_batch_job(
...
-        batch_request = batch_client.get_request(batch_request)
+    with tqdm(total=100, initial=progress, desc="Completion percentage") as progress_bar:
+        while progress < 100:
+            time.sleep(sleep_time)
 
-    batch_request.raise_for_status(status=[BatchRequestStatus.FAILED, BatchRequestStatus.CANCELED])
-    return batch_request
+            new_progress = current_completion_percentage()
+            progress_bar.update(new_progress - progress)
+            progress = new_progress
```
AleksMat, line 235: The function could actually return the payload of the status endpoint response.

sentinelhub/api/batch/utils.py
```diff
@@ -86,39 +214,28 @@ def monitor_batch_job(
...
+    progress = current_completion_percentage()
 
-    batch_request = batch_client.get_request(batch_request)
-    while batch_request.status in [BatchRequestStatus.CREATED, BatchRequestStatus.ANALYSING]:
-        LOGGER.info("Batch job has a status %s, sleeping for %d seconds", batch_request.status.value, sleep_time)
-        time.sleep(sleep_time)
-        batch_request = batch_client.get_request(batch_request)
+    with tqdm(total=100, initial=progress, desc="Completion percentage") as progress_bar:
+        while progress < 100:
+            time.sleep(sleep_time)
```
AleksMat, line 230: This time could probably be by default around 5-10 seconds I guess. I actually recommend checking with Marjan about this.

sentinelhub/api/batch/utils.py
zigaLuksic: yes, this was also one of the reasons i was considering splitting

sentinelhub/api/batch/utils.py
zigaLuksic: i packed it into a functions so that the code is easier to parse. I'd prefer `current_completion = lambda: ...` but linters don't like that.

sentinelhub/api/batch/utils.py
```diff
@@ -117,18 +217,34 @@ def monitor_batch_analysis(
...
+    request_status = BatchRequestStatus(batch_client.get_status(batch_request)["status"])
+    while request_status in [BatchRequestStatus.CREATED, BatchRequestStatus.ANALYSING]:
+        LOGGER.info("Batch job has a status %s, sleeping for %d seconds", request_status.value, sleep_time)
+        time.sleep(sleep_time)
+        request_status = BatchRequestStatus(batch_client.get_status(batch_request)["status"])
 
-    return tiles_per_status
+    batch_request = batch_client.get_request(batch_request)
+    batch_request.raise_for_status(status=[BatchRequestStatus.FAILED, BatchRequestStatus.CANCELED])
+    return batch_request
```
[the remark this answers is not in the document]
zigaLuksic: If we ever want to unify the two functions then it makes sense to return the full request.

sentinelhub/api/batch/utils.py
```diff
@@ -86,8 +110,85 @@ def monitor_batch_job(
...
+    batch_client = SentinelHubBatchStatistical(config=config)
+
+    request_status = batch_client.get_status(batch_request)
+    progress = request_status["completionPercentage"]
+    with tqdm(total=100, initial=progress, desc="Completion percentage") as progress_bar:
+        while progress < 100:
+            time.sleep(sleep_time)
+
+            request_status = batch_client.get_status(batch_request)
+            progress_bar.update(request_status["completionPercentage"] - progress)
```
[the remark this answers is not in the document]
zigaLuksic: floats work :)

tests/api/batch/test_utils.py
```diff
@@ -114,12 +117,70 @@ def _tile_status_counts_to_tiles(tile_status_counts: Dict[BatchTileStatus, int])
...
+@pytest.mark.parametrize(
+    "progress_sequence",
+    [
+        (0, 10, 30.5, 70, 90, 99, 100),
+        (50.712, 80, 100),
+        (100,),
+    ],
+)
+@pytest.mark.parametrize("config", [SHConfig(), None])
+@pytest.mark.parametrize("sleep_time", [60, 1000])
```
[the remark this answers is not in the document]
zigaLuksic: done

sentinelhub/api/batch/utils.py
```diff
@@ -86,8 +110,85 @@ def monitor_batch_job(
...
+            progress = request_status["completionPercentage"]
```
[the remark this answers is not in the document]
zigaLuksic: i would've preferred to extract `batch_client.get_status(batch_request)["completionPercentage"]` to a function, but i guess `new_progress` is a middle ground that we both like. Done

sentinelhub/api/batch/utils.py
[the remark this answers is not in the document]
AleksMat: Ok, then let's leave this one extra request.

PR #312, opened 2022-08-10
Utility for downloading Statistical Batch API results

AleksMat:
A class to download results from an S3 bucket and either return or save them locally. It is another implementation of `DataRequest` class. Besides that I also had to generalize and fix a few things in `AwsDownloadClient` to make this implementation possible.

The following will be done in next PRs:

- unit tests for `AwsBatchResults`,
- a method of `AwsBatchResults` that will be returning results in form of a pandas dataframe (maybe in the next package version?)

### DIFF
--- a/sentinelhub/aws/batch.py
+++ b/sentinelhub/aws/batch.py
@@ -1,0 +1,88 @@
+"""
+Module implementing utilities for collecting data, produced with Sentinel Hub Statistical Batch API, from an S3 bucket.
+"""
+from typing import Any, List, Optional, Sequence, Union
+
+from ..api.batch.statistical import BatchStatisticalRequest, SentinelHubBatchStatistical
+from ..base import DataRequest
+from ..config import SHConfig
+from ..constants import MimeType
+from ..download.models import DownloadRequest
+from .client import AwsDownloadClient
+
+
+class AwsBatchResults(DataRequest):
+    """A utility class for downloading results of Batch Statistical API from an S3 bucket."""
+
+    def __init__(
+        self,
+        batch_request,
+        *,
+        feature_ids: Optional[Sequence[Union[str, int]]] = None,
+        data_folder: Optional[str] = None,
+        config: Optional[SHConfig] = None
+    ):
+        """
+        :param batch_request: Info about a batch request - either an instance of `BatchStatisticalRequest` or a
+            batch ID or a raw payload of the batch response.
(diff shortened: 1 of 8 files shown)


sentinelhub/aws/batch.py
```diff
@@ -0,0 +1,90 @@
...
+        config: Optional[SHConfig] = None,
+    ):
+        """
+        :param batch_request: Info about a batch request - either an instance of `BatchStatisticalRequest` or a
+            batch ID or a raw payload of the batch response.
+        :param feature_ids: A list of feature IDs of saved results on the bucket. If provided it will download only
+            these results. If not provided it will collect the names of all JSON files from results folder on the
+            bucket and download all of them. Note that it is recommended that you provide this parameter otherwise this
+            class will have to make additional requests to the S3 bucket in order to list all features from the folder.
+        :param data_folder: A location of the directory where the downloaded data could be saved.
```
zigaLuksic, line 32:
```suggestion
        :param data_folder: Directory to which the files should be saved.
```

sentinelhub/aws/batch.py
```diff
@@ -0,0 +1,90 @@
...
+        :param config: A config object that contains AWS credentials to access the S3 bucket with results.
```
zigaLuksic, line 33: i was thinking if the user could perhaps just specify the AWS credentials (or use an `S3Specification` object) as they are likely still around from the request creation, but I guess it's not that much hassle to adjust an `SHConfig` :thinking:

sentinelhub/base.py
```diff
@@ -24,7 +24,7 @@ class DataRequest(metaclass=ABCMeta):
     """
 
     def __init__(
-        self, download_client_class: type, *, data_folder: Optional[str] = None, config: Optional[SHConfig] = None
+        self, download_client_class: Callable, *, data_folder: Optional[str] = None, config: Optional[SHConfig] = None
```
zigaLuksic, line 27: good call

sentinelhub/aws/batch.py
AleksMat: I suggest we go with `config` style because we are using it already throughout this package. This way we can have only one parameter that contains AWS and SH credentials. (Note that SH credentials here might also be required if user doesn't provide an instance of `BatchStatisticalRequest`.)

zigaLuksic (approved):

PR #319, opened 2022-08-26
Edit stat_to_df helper function in batch_statistical example notebook

chorng:
Main changes:
- `is_valid_entry` logic: Add to dataframe if there's at least one valid band as `dataMask` can be different between outputs.
- Add `result_to_dataframe`: Take the return of `get_data` as input, generate the dataframe and collect the geometries that failed on request.
- Add `_histogram_to_dataframe`: Transform stat api histogram into a dataframe. The histogram is stored in two columns, bins and counts, as lists to make plotting easier.


sentinelhub/data_utils.py
```diff
@@ -0,0 +1,166 @@
...
+            if df_entry:
+                df_entry["interval_from"] = parse_time(interval["interval"]["from"])
+                df_entry["interval_to"] = parse_time(interval["interval"]["to"])
+                df_entries.append(df_entry)
+
+    return df_entries
+
+
+def statistical_to_dataframe(
+    result_data: List[JsonDict], excl_stats: Optional[List[str]] = None, incl_hist: bool = False
```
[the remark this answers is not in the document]
zigaLuksic: Also write a description of what this does. `Flag to transform histogram` is not very informative. In fact the entire docstring needs to be improved.

sentinelhub/data_utils.py
```diff
@@ -0,0 +1,166 @@
...
+Module with statistics to dataframe transformation.
+"""
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+from .exceptions import PandasNotInstalled
+from .time_utils import parse_time
+from .type_utils import JsonDict
+
+
+def _extract_hist(hist_cols: Tuple[str, str], hist_data: List[JsonDict]) -> Dict[str, List[float]]:
```
zigaLuksic, line 11: why is `hist_cols` a tuple of strings instead of two parameters? It would probably even be best if the function returns two values (bins and counts) and they are transformed into DF form in the outside function (better typechecking)

sentinelhub/data_utils.py
```diff
@@ -0,0 +1,166 @@
...
+    for idx in range(nbins):
+        bins.append(hist_data[idx]["lowEdge"])
+        if idx == nbins - 1:
+            bins.append(hist_data[idx]["highEdge"])
+        counts.append(hist_data[idx]["count"])
+
+    return {hist_cols[0]: bins, hist_cols[1]: counts}
+
+
+def _extract_percentiles(col_name: str, perc_data: Dict[str, float]) -> Dict[str, float]:
```
zigaLuksic, line 30: `perc_data` shouldn't be abreviated

sentinelhub/data_utils.py
```diff
@@ -0,0 +1,166 @@
...
+
+    :param col_name: Output name and band name. Example: ndvi_B0.
+    :param perc_data: An input representation of percentile data of <col_name>.
+    :return: Percentile data as a pandas.DataFrame entry. Examples: {"ndvi_B0_<perc>": perc_val}.
+    """
+    perc_entry = {}
+    for perc, perc_val in perc_data.items():
+        perc_col_name = f"{col_name}_{perc}"
+        perc_entry[perc_col_name] = perc_val
+    return perc_entry
```
zigaLuksic, line 41: could this perhaps be just `return {f"{col_name}_{perc}": value for perc, value in perc_data.items()}` If so this function is not necessary.

sentinelhub/data_utils.py
```diff
@@ -0,0 +1,166 @@
...
+    :param interval_output: An input representation of statistics of an aggregation interval.
+    :param excl_stats: Unwanted statistical name.
+    :param incl_histogram: Flag to transform histogram.
+    :return: Statistics as a pandas.DataFrame entry.
+    """
+    stat_entry: Dict[str, Union[List[float], float]] = {}
+    for output_name, output_data in interval_output.items():
+        for band_name, band_values in output_data["bands"].items():
+            band_stats = band_values["stats"]
+            if band_stats["sampleCount"] == band_stats["noDataCount"]:
```
zigaLuksic, line 58: add a comment explaining why this results in a `break`

zigaLuksic (changes requested):

sentinelhub/data_utils.py
```diff
@@ -0,0 +1,154 @@
...
+            df_entry: Dict[str, Any] = _extract_stats(interval["outputs"], exclude_stats)
+            if df_entry:
+                df_entry["interval_from"] = parse_time(interval["interval"]["from"])
+                df_entry["interval_to"] = parse_time(interval["interval"]["to"])
+                df_entries.append(df_entry)
+
+    return df_entries
+
+
+def statistical_to_dataframe(result_data: List[JsonDict], exclude_stats: Optional[List[str]] = None) -> Any:
```
chorng, line 92: Rethink the need of `include_histograms`. The histogram data is in the result only if it is requested in the payload. Doesn't make sense to skip it when the data is there.

sentinelhub/data_utils.py
```diff
@@ -0,0 +1,154 @@
...
+                        stat_entry.update(
+                            {
+                                f"{col_name}_{percentile_name}": percentile_value
+                                for percentile_name, percentile_value in value.items()
+                            }
+                        )
+                    else:
+                        stat_entry[col_name] = value
+
+                if "histogram" in band_values:
```
chorng, line 59: Apply the same logic as `percentiles`

sentinelhub/data_utils.py
```diff
@@ -3,17 +3,18 @@
 """
 from typing import Any, Dict, List, Optional, Tuple, Union
 
-from .exceptions import PandasNotInstalled
 from .time_utils import parse_time
 from .type_utils import JsonDict
 
+_PANDAS_IMPORT_MESSAGE = "The Pandas backend is not installed by default. We suggest you install the `pandas` package."
```
zigaLuksic, line 9: Change to something along the lines of `"To use this function you need to install the `pandas` library, which is not a requirement of `sentinelhub-py`." The docstring of the functions should also mention that

sentinelhub/data_utils.py
```diff
@@ -99,35 +89,33 @@ def _extract_response_data(
...
+def statistical_to_dataframe(result_data: List[JsonDict], exclude_stats: Optional[List[str]] = None) -> Any:
+    """Transform (Batch) Statistical API results into a pandas.DataFrame
 
-    :param result_data: An input representation of (Batch) Statistical API result.
-    :param excl_stats: Unwanted statistical name.
-    :param incl_hist: Flag to transform histogram.
+    :param result_data: An input representation of (Batch) Statistical API result returned from
+        `AwsBatchResults.get_data()`. Each JsonDict in the list is a Statistical API response of an input geometry.
+    :param exclude_stats: Statistical API has `min`, `max`, `mean`, and `stDev` as default statistics.
+        The default statistics defined in this parameter will be removed from DataFrame.
```
zigaLuksic, line 98: I am not sure this description is accurate

sentinelhub/data_utils.py
```diff
@@ -24,73 +25,62 @@ def _extract_hist(hist_cols: Tuple[str, str], hist_data: List[JsonDict]) -> Dict
...
+                if stat_name not in exclude_stats:
                     col_name = f"{output_name}_{band_name}_{stat_name}"
                     if stat_name == "percentiles":
-                        stat_entry.update(_extract_percentiles(col_name, value))
+                        stat_entry.update(
+                            {
+                                f"{col_name}_{percentile_name}": percentile_value
+                                for percentile_name, percentile_value in value.items()
+                            }
+                        )
```
zigaLuksic, line 55: Would this be shorter if you did it manually (without update), e.g.
```python
for percentile_name, percentile_value in value.items():
    stat_entry[f"..."] = percentile_value
```
?

The same should be tried later for histogram values

sentinelhub/data_utils.py
```diff
@@ -24,73 +25,62 @@ def _extract_hist(hist_cols: Tuple[str, str], hist_data: List[JsonDict]) -> Dict
...
                     else:
                         stat_entry[col_name] = value
 
-                if incl_histogram:
+                if "histogram" in band_values:
                     band_bins = band_values["histogram"]["bins"]
-                    hist_col_names = (f"{output_name}_{band_name}_bins", f"{output_name}_{band_name}_counts")
-                    stat_entry.update(_extract_hist(hist_col_names, band_bins))
+                    hist_bins_value, hist_counts_value = _extract_hist(band_bins)
```
zigaLuksic, line 61: names could be `hist_bins` and `hist_counts`

zigaLuksic (changes requested):

sentinelhub/data_utils.py
chorng: Would something like `"The statistic names defined in this parameter will not be included in the DataFrame."`? This argument will be mostly used in case of the default statistics are not interested by users.

sentinelhub/data_utils.py
chorng: If a for loop is added here there's a complain of `too many nested clocks (6/5)` from mypy. I also think the original styling in the [helper function](https://sentinelhub-py.readthedocs.io/en/latest/examples/statistical_request.html#Helper-function) is better. 

Any suggested workaround in this case?

zigaLuksic: I would prefer disabling the linter with an inline comment (it's probably pylint, so something along `# pylint: disable=<error-code>`)
because this way you have more indentation and a for loop remains.

sentinelhub/data_utils.py
zigaLuksic: your suggestion sounds better, yes

zigaLuksic:
Now please re-run the notebook with the final helper function.

chorng:
> Now please re-run the notebook with the final helper function.

Why `parse_time` can be imported directly from `sentinelhub` but `statistical_to_dataframe`?

chorng:
> > Now please re-run the notebook with the final helper function.
> 
> Why `parse_time` can be imported directly from `sentinelhub` but `statistical_to_dataframe`?

Sorry, it needs to be imported in the `__init__.py` file :see_no_evil:

sentinelhub/__init__.py
```diff
@@ -54,6 +54,7 @@
 )
 from .data_collections import DataCollection
 from .data_collections_bands import Band, Unit
+from .data_utils import get_failed_requests, statistical_to_dataframe
```
zigaLuksic, line 57: I am not sure we'd want to add this to the global `__init__`, especially since it requires pandas. I think it's completely fine if users have to import it directly with `from sentinelhub.data_utils import ...`

sentinelhub/__init__.py
```diff
@@ -54,7 +54,7 @@
 )
 from .data_collections import DataCollection
 from .data_collections_bands import Band, Unit
-from .data_utils import get_failed_statistical_requests, statistical_to_dataframe
+from .data_utils import get_failed_statistical_requests
```
zigaLuksic, line 57: was this one left in intentionally? I'd perhaps leave it out as well for now, and if it sees a lot of use we can maybe add it in later. It's much harder to remove it later (because importing from sentinelhub.data_utils will remain valid even if we add it to sentinelhub)

chorng: Ok, I thought it was not using pandas so I left it in. I'll remove it from `__init__.py`

zigaLuksic (approved):

PR #323, opened 2022-08-31
add pull_request trigger to CI

zigaLuksic:


.github/workflows/ci_action.yml
```diff
@@ -66,6 +82,8 @@ jobs:
         run: pylint sentinelhub
 
   test-on-github:
+    needs: check_duplication
+    if: ${{ needs.check_duplication.outputs.should_skip != 'true' }}
```
AleksMat, line 86: Running integration tests for a PR from an external contributor can be a security issue.

The code has access to secret variables, including SH and AWS credentials. This means that a malicious contributor could modify the code, which would be run by CI, in a way that it would send credentials to them.

A better solution for this problem would be that for any external PR someone from SH team would have to manually approve that CI can be run for that PR. I think something like that might be possible but I'm not exactly sure.

zigaLuksic: That is set in the settings (unless i read the [documentation](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) wrong). By default it requires approval only for first-time contributors, but I strengthened it to any outside contributor.

AleksMat: Oh nice, then it looks good. :+1: 

Cheers :sunglasses: :wave:

zigaLuksic:
will re-open a cleaned version

PR #329, opened 2022-09-18
Added types in tests

AleksMat:
Another thing I wanted to finish before I left - to add type hints to all tests.

On one hand it seems redundant to have e.g. a return type ` -> None:` in each test but a general benefit of having type hints is that now if you run mypy on tests it will show a wide range of minor typing issues. Some of them could just be suppressed but some of them actually suggest beneficial improvements in the package. So in the future, you can try to improve some of them.


zigaLuksic (approved):
I checked the non-test files, and scanned the rest. It looks great, thanks a bunch :heart:

PR #330, opened 2022-09-21
Adjust the catalog client to use the catalog 1.0.0

zigaLuksic:
The main changes are to the querying, since now we use the `filter` parameter in CQL2 instead.

The tests were also changed, because the old tests did a very poor job (if you just disabled querying/filtering they still all passed).

### DIFF
--- a/sentinelhub/api/catalog.py
+++ b/sentinelhub/api/catalog.py
@@ -2,7 +2,8 @@
 A client interface for `Sentinel Hub Catalog API <https://docs.sentinel-hub.com/api/latest/api/catalog>`__.
 """
 import datetime as dt
-from typing import Any, Iterable, List, Optional, Union
+import sys
+from typing import Any, Dict, Iterable, List, Optional, Union
 
 from ..base import FeatureIterator
 from ..data_collections import DataCollection, OrbitDirection
@@ -11,6 +12,11 @@ from ..time_utils import parse_time, parse_time_interval, serialize_time
 from ..type_utils import JsonDict, RawTimeIntervalType, RawTimeType
 from .base import SentinelHubService
 from .utils import remove_undefined
+
+if sys.version_info < (3, 8):
+    from typing_extensions import Literal
+else:
+    from typing import Literal  # pylint: disable=ungrouped-imports
 
 
 class SentinelHubCatalog(SentinelHubService):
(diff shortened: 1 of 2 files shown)


sentinelhub/api/catalog.py
```diff
@@ -106,13 +114,23 @@ def search(
         :param geometry: A search geometry, it will always reprojected to WGS 84 before being sent to the service.
             This parameter is defined with parameter `intersects` at the service.
         :param ids: A list of feature ids as defined in service documentation
-        :param query: A STAC query described in Catalog API documentation
+        :param filter: A STAC filter in CQL2, described in Catalog API documentation
+        :param filter_lang: How to parse CQL2 of the `filter` input, described in Catalog API documentation
```
veseln, line 118: If there is any openly available documentation, I would add a link somewhere in the documentation. Because currently it is not clear for somebody not familiar.

sentinelhub/api/catalog.py
```diff
@@ -146,35 +162,68 @@ def search(
...
+        if filter_lang == "cql2-text":
+            if not (filter_query is None or isinstance(filter_query, str)):
+                raise ValueError(input_missmatch_msg)
+
+            text_queries = [f"{field}='{value}'" for field, value in collection_filters.items()]
+            if filter_query:
+                text_queries.append(filter_query)
+            return " AND ".join(text_queries) if text_queries else None
+
+        # filter_lang == cql2-json
```
veseln, line 192: what happens if something else gets passed by the user as `filter_lang`, if I understand correctly, then this would be executed as `cql2-json`? 

If I understand this correctly, than either some assert is needed to check that the value is one of these,  or this needs to be rewritten as something like: 

```
if filter_lang == "cql2-text"
  ...
elif filter_lang == "cql2-json"
  ...
else:
  raise ValueError("Unsupported filter-lang")

```

sentinelhub/api/catalog.py
zigaLuksic: I talked with Majcen, he said he didn't find anything useful on CQL2

sentinelhub/api/catalog.py
veseln: I think its best to assume that the user doesn't have type checker and has fat fingers 🤣

sentinelhub/api/catalog.py
veseln: 🤷‍♂️  every man and woman for themselves then

PR #339, opened 2022-11-02
Remove os_utils module and fix deprecations

jgersak:


sentinelhub/io_utils.py
```diff
@@ -166,6 +165,13 @@ def read_numpy(filename: str) -> np.ndarray:
     return np.load(filename)
 
 
+def _create_parent_folder(filename: str) -> None:
+    """Create parent folder for input filename recursively"""
```
zigaLuksic, line 169: what is recursive about this?

jgersak: `os.makedirs` is recursive directory creation function that makes all intermediate-level directories needed to contain the leaf directory 
docstring could be removed add almost zero meaning

zigaLuksic: yes, in this case it's more confusing than explanatory.

jgersak: removed

PR #340, opened 2022-11-04
Update test io_utils 

jgersak:
add test for txt and xml
make jpg test for himself

### DIFF
--- a/tests/test_io_utils.py
+++ b/tests/test_io_utils.py
@@ -1,9 +1,10 @@
 import os
-from platform import python_implementation
 from typing import Tuple
+from xml.etree import ElementTree as ET
 
 import numpy as np
 import pytest
+from fs.tempfs import TempFS
 from pytest import approx
 
 from sentinelhub import read_data, write_data
@@ -14,7 +15,6 @@ from sentinelhub.exceptions import SHUserWarning
     "filename, mean, shape",
     [
         ("img.tif", 13577.494856, (2048, 2048, 3)),
-        ("img.jpg", 52.412425, (2048, 2048, 3)),
         ("img.png", 52.33736, (2048, 2048, 3)),
         ("img-8bit.jp2", 47.09060, (343, 343, 3)),
         ("img-15bit.jp2", 0.3041897, (1830, 1830)),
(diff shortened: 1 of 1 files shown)


tests/test_io_utils.py
```diff
@@ -14,35 +15,67 @@
...
         ("img-15bit.jp2", 0.3041897, (1830, 1830)),
         ("img-16bit.jp2", 0.3041897, (1830, 1830)),
     ],
 )
 def test_img_read(input_folder: str, output_folder: str, filename: str, mean: float, shape: Tuple[int, ...]) -> None:
-    is_jpeg = filename.endswith("jpg")
     img = read_data(os.path.join(input_folder, filename))
 
+    if filename.endswith("txt"):
+        img = np.fromstring(img)
```
zigaLuksic, line 28: there is no test case where the filename ends with `.txt`
also no reason for the data to be an image in that case, just use a human readable string. Good luck finding the difference between two 40.000 character strings (where one symbol changed due to encoding)

tests/test_io_utils.py
```diff
@@ -14,35 +15,67 @@
...
+
     assert img.shape == shape
+    assert np.mean(img) == approx(mean, abs=1e-4)
+    assert img.flags["WRITEABLE"], "Obtained numpy array is not writeable"
+
+    output_file_path = os.path.join(output_folder, filename)
+    write_data(output_file_path, img)
+    new_img = read_data(output_file_path)
+
+    assert np.array_equal(img, new_img), "Original and saved image are not the same"
```
zigaLuksic, line 38: why are we testing the write method inside a read test?
either the test name should be changed, or better yet, we should have a separate test for the write+read

tests/test_io_utils.py
```diff
@@ -14,35 +15,67 @@
...
+        file_path = os.path.join(filesystem.root_path, "test-string.txt")
+
+        text = "sentinelhub-py is often shortened to sh-py"
+        write_data(file_path, text)
+        assert text == read_data(file_path)
+
+
+def test_read_write_xml() -> None:
+    with TempFS() as filesystem:
+        file_path = os.path.join(filesystem.root_path, "test-xml.xml")
```
zigaLuksic, line 68: I believe `filesystem.getsyspath` would be nicer here.

tests/test_io_utils.py
```diff
@@ -14,35 +15,67 @@
...
+        data = ET.SubElement(root, "data")
+        ET.SubElement(data, "field1", name="BANDS-S2-L1C").text = "some value1"
+        ET.SubElement(data, "field2", name="CLP").text = "some value2"
+        tree = ET.ElementTree(root)
 
-    if not is_jpeg:
-        assert np.array_equal(img, new_img), "Original and saved image are not the same"
+        write_data(file_path, tree)
+        tree_new = read_data(file_path)
+        assert set(tree_new.getroot().itertext()) == set(tree.getroot().itertext())
```
zigaLuksic, line 78: The first step would be to check if its the instance of `ET.ElementTree`. Same goes for all the tests above, check that the type of read object is correct.
does this detect the hierarchy correctly? I know it's not our job to check the library, but this equality test seems odd

tests/test_io_utils.py
jgersak: removed

tests/test_io_utils.py
jgersak: done

tests/test_io_utils.py
zigaLuksic: yes

jgersak: done

tests/test_io_utils.py
```diff
@@ -14,35 +15,61 @@
...
 
-    if is_jpeg or python_implementation() != "PyPy":
-        assert np.mean(img) == approx(mean, abs=1e-4)
 
-    assert img.flags["WRITEABLE"], "Obtained numpy array is not writeable"
+xml_root = ET.Element("EOPatch")
+xml_data = ET.SubElement(xml_root, "data")
+ET.SubElement(xml_data, "field1", name="BANDS-S2-L1C").text = "some value1"
+ET.SubElement(xml_data, "field2", name="CLP").text = "some value2"
+TEST_XML = ET.ElementTree(xml_root)
```
zigaLuksic, line 37: It would be great to have this in a fixture, but as you probably noticed, fixtures and `parametrize` don't mix. UNLESS you use the [lazy-fixture](https://github.com/tvorog/pytest-lazy-fixture) plugin. We already use it in eo-learn and I see no reason not to use it here as well (it is just a develop dependency, so it doesn't bloat the package).

Please add it to the requirements and change the above test case into a fixture, as it should improve the readability.

tests/test_io_utils.py
```diff
@@ -14,35 +15,61 @@
...
+        ("img.tif", np.arange(5 * 5 * 3).reshape((5, 5, 3))),
+        ("img.png", np.arange((5 * 5 * 3), dtype=np.uint8).reshape((5, 5, 3))),
+        ("img-8bit.jp2", np.arange((5 * 5 * 3), dtype=np.uint8).reshape((5, 5, 3))),
+        ("img-15bit.jp2", np.arange((5 * 5 * 3), dtype=np.uint8).reshape((5, 5, 3))),
+        ("img-16bit.jp2", np.arange((5 * 5 * 3), dtype=np.uint8).reshape((5, 5, 3))),
+        ("test-string.txt", "sentinelhub-py is often shortened to sh-py"),
+        ("test-xml.xml", TEST_XML),
+    ],
+)
+def test_write(filename: str, data: Union[str, np.ndarray, ET.ElementTree]) -> None:
```
zigaLuksic, line 52: i'd perhaps rename it to `test_write_read` :thinking: 
Also do you think it could be worthwhile to check `filesystem.exists(filename)`?

jgersak: the thing is that this check checks working of fs.tempfs 
and is kinda move in other direction from micro tests
but i can add it it one line of code that's relay fast

tests/test_io_utils.py
```diff
@@ -30,11 +31,13 @@ def test_img_read(input_folder: str, filename: str, mean: float, shape: Tuple[in
     assert img.flags["WRITEABLE"], "Obtained numpy array is not writeable"
 
 
-xml_root = ET.Element("EOPatch")
-xml_data = ET.SubElement(xml_root, "data")
-ET.SubElement(xml_data, "field1", name="BANDS-S2-L1C").text = "some value1"
-ET.SubElement(xml_data, "field2", name="CLP").text = "some value2"
-TEST_XML = ET.ElementTree(xml_root)
+@pytest.fixture
+def xml_test():
```
zigaLuksic, line 35: i'd go with `xml_testcase` or `xml_data`, just so it doesn't carry a connotation of being a test itself

jgersak: done

tests/test_io_utils.py
jgersak: done

Issue #354, opened 2022-12-08
[FEAT] Enable BBoxSplitter to accept bbox_size parameter


ColinMoldenhauer:
**What is the problem? Please describe.**

Opposed to the UTM splitters, the [`BBoxSplitter`](https://sentinelhub-py.readthedocs.io/en/latest/reference/sentinelhub.areas.html#sentinelhub.areas.BBoxSplitter) does not accept a parameter `bbox_size`, but only a `split_shape`, splitting the AOI into *n x m* equally sized boxes.
This limitation makes obtaining boxes of a certain pixel size unnecessarily complicated.

**Here's the solution**
Below sample solution is basically a copy of the existing `BBoxSplitter`, the only change being the call 
```python
self.area_bbox.get_partition(size_x=size_x, size_y=size_y)
```
instead of 
```python
bbox_partition = self.area_bbox.get_partition(num_x=columns, num_y=rows)
```
as `get_partition()` already implements the necessary functionality to realize this feature request.

**Sample solution:**
```python
class BBoxSplitterSize(AreaSplitter):
    """A tool that splits the given area into smaller parts. Given the area it calculates its bounding box and splits
    it into smaller bounding boxes of equal size. Then it filters out the bounding boxes that do not intersect the
    area. If specified by user it can also reduce the sizes of the remaining bounding boxes to best fit the area.
    """

    def __init__(
        self,
        shape_list: Iterable[Union[Polygon, MultiPolygon, _BaseGeometry]],
        crs: CRS,
        split_shape: Union[int, Tuple[int, int]],
        **kwargs: Any,
    ):
        """
        :param shape_list: A list of geometrical shapes describing the area of interest
        :param crs: Coordinate reference system of the shapes in `shape_list`
        :param split_shape: Parameter that describes the size of patches (in meters) into which the area bounding box will be split.
            It can be a tuple of the form `(width, height)` which means the area bounding box will be split into patches
            of size (width, height). It can also be a single integer `size` which is the same as `(size, size)`.
        :param reduce_bbox_sizes: If `True` it will reduce the sizes of bounding boxes so that they will tightly fit
            the given area geometry from `shape_list`.
        """
        self.split_shape = _parse_to_pair(split_shape, allowed_types=(int,), param_name="split_shape")
        super().__init__(shape_list, crs, **kwargs)

    def _make_split(self) -> Tuple[List[BBox], List[Dict[str, object]]]:
        size_x, size_y = self.split_shape
[...]

PR #357, opened 2022-12-16
BBoxSplitter: Split by Size

ColinMoldenhauer:
Solves [Issue 354](https://github.com/sentinel-hub/sentinelhub-py/issues/354) by
- making `split_shape` optional
- introducing the optional parameter `split_size`
- allowing only one of the above parameters to be set


sentinelhub/areas.py
```diff
@@ -193,15 +194,34 @@ def __init__(
...
         :param reduce_bbox_sizes: If `True` it will reduce the sizes of bounding boxes so that they will tightly fit
             the given area geometry from `shape_list`.
         """
-        self.split_shape = _parse_to_pair(split_shape, allowed_types=(int,), param_name="split_shape")
+        if (split_shape is not None) and (split_size is None):
+            self.split_shape = _parse_to_pair(split_shape, allowed_types=(int,), param_name="split_shape")
+            self.split_size = None
+        elif (split_shape is None) and (split_size is not None):
+            self.split_shape = None
+            self.split_size = _parse_to_pair(split_size, allowed_types=(int,), param_name="split_shape")
```
zigaLuksic, line 209: do we have a sexier way of doing this?

I'd try to condense this information into an attribute of type
```python
self.split_params: Tuple[Literal["shape", "size"], Tuple[int, int]]
```

But it could be that the code gets uglier this way.

zigaLuksic (commented):
Could you perhaps add an additional test for this?

sentinelhub/areas.py
ColinMoldenhauer: I like the idea, but it doesn't really get prettier

ColinMoldenhauer:
> Could you perhaps add an additional test for this?

I am afraid I am not familiar with the best practices of writing tests.

zigaLuksic:
> > Could you perhaps add an additional test for this?
> 
> I am afraid I am not familiar with the best practices of writing tests.

No problem, the test will be added in a separate MR :)
I see that the refactoring got rid of a mypy error (yay, step in right direction)
the only thing left is to format the code, apparently `black` doesn't agree with something.

To do this, you need to install `black` (can be via pip) and then run it on the file. It's best if you call it from the root of the repo, so that our configuration from `pyproject.toml` will be used, so `black sentinelhub/areas.py`.

sentinelhub/areas.py
zigaLuksic: i like it this way, lets leave it in this shape

zigaLuksic:
@ColinMoldenhauer  Thank you! The contribution is very appreciated :D

PR #365, opened 2023-01-09
Add dataFilter defaults of {} to batch statistical

zigaLuksic:
The regular statistical API already did that. However it did that in a mutable way, which we now avoid by copying elements that are to be adjusted.


chorng:
@zigaLuksic Tested for normal stats and batch stats and both worked

chorng (approved):

PR #377, opened 2023-01-24
Add tests for data_utils

chorng:


tests/test_data_utils.py
```diff
@@ -0,0 +1,43 @@
...
+    "ndvi_B0_bins": list,
+    "ndvi_B0_counts": list,
+    "interval_from": dt.date,
+    "interval_to": dt.date,
+    "identifier": str,
+}
+
+
+def test_statistical_to_dataframe() -> None:
+    batch_stat_results = read_data(os.path.join(os.path.dirname(__file__), "TestInputs", "batch_stat_results.json"))
```
zigaLuksic, line 26: use the `input_folder` fixture for the path to the input folder (like [here](https://github.com/sentinel-hub/sentinelhub-py/blob/master/tests/test_decode.py#L11))

tests/test_data_utils.py
```diff
@@ -0,0 +1,43 @@
...
+    "ndvi_B0_stDev": float,
+    "ndvi_B0_sampleCount": np.int64,
+    "ndvi_B0_noDataCount": np.int64,
+    "ndvi_B0_percentiles_50.0": float,
+    "ndvi_B0_bins": list,
+    "ndvi_B0_counts": list,
+    "interval_from": dt.date,
+    "interval_to": dt.date,
+    "identifier": str,
+}
```
zigaLuksic, line 22: rewrite so that it is a sequence of pairs like `(float, ["ndvi_B0_min", "ndvi_B0_max", "ndvi_B0_mean", ...])`

tests/test_data_utils.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_statistical_to_dataframe() -> None:
+    batch_stat_results = read_data(os.path.join(os.path.dirname(__file__), "TestInputs", "batch_stat_results.json"))
+    df = statistical_to_dataframe(batch_stat_results)
+    num_polys = len(set(df["identifier"]))
+    num_columns = len(df.columns)
+    num_valid_rows = len(df)
+    assert num_polys == 2
+    assert num_columns == 12
+    assert num_valid_rows == 2
```
zigaLuksic, line 33: you can do `assert len(set(df["identifier"])) == 2, "Wrong number of polygons"` to get rid of multiple lines.
By using`assert cond, msg`, if the assert fails the `msg` will be displayed.

tests/test_data_utils.py
```diff
@@ -0,0 +1,43 @@
...
+    for column, data_type in column_types.items():
+        assert isinstance(df[column].iloc[0], data_type)
+
+
+def test_get_failed_statistical_requests() -> None:
+    batch_stat_failed_results = read_data(
+        os.path.join(os.path.dirname(__file__), "TestInputs", "batch_stat_failed_results.json")
+    )
```
zigaLuksic, line 41: the test case here is really minimal, so you can just add it to the `.py` file as a dictionary. I would also like a case where the list is empty, and I'd extend this case to have two or three requests in it.

chorng: Two more cases will be added, but they will cause an error in `get_failed_statistical_request` function. @Adrok24 suggested to first added the simplest case that works with the current function, then I fix the function and add other test cases in another commit.

zigaLuksic: what if you add the two cases, but comment them out for now (with an additional comment that the function has to be fixed)?
Perhaps even better: a [skip](https://docs.pytest.org/en/7.1.x/how-to/skipping.html) for those tests

zigaLuksic (approved):

PR #380, opened 2023-01-27
minor adjustments found while code-surfing

zigaLuksic:


sentinelhub/download/rate_limit.py
```diff
@@ -121,9 +109,7 @@ def get_wait_time(
         expected_content = max(self.content + elapsed_time * self.refill_per_second - overall_completed_cost, 0)
 
         if self.is_fixed():
-            if expected_content < cost_per_request:
-                return -1
-            return 0
+            return -1 if expected_content < cost_per_request else 0
```
jgersak, line 112:
```suggestion
            return -1 * int(expected_content < cost_per_request)
```
do not know why i do not like `if` in return ... this dose not look much better :(

zigaLuksic: i am very against casting bools to integers unless absolutely needed

PR #383, opened 2023-01-31
Fix CRS class to support fiona 1.9.0

zigaLuksic:
Also adjust types in testing_utils to not waste another MR for such a minor thing


jgersak (approved):
we probably need more direct (encapsulated) test of CRSMeta but not part of this MR

PR #388, opened 2023-02-01
Refactor config tests, try to improve security

zigaLuksic:
Currently there is a small possibility of credentials being revealed in the tests if a test fails.

With this I introduce a fixture that caches the current config.json, and restores it after the tests. The first test (on which all other depend) checks that the config no longer contains credentials. If it still does, the test suite is aborted.


jgersak (approved):
LGTM, really like use of `pytest.mark.dependency` .

PR #391, opened 2023-02-02
Remove BBoxCollection

jgersak:
update areas (replaced with List[BBox])
update tests accordingly


sentinelhub/areas.py
```diff
@@ -427,15 +427,13 @@ def __init__(
...
-            return bbox_grid
+    def _parse_bbox_grid(bbox_grid: List[BBox]) -> List[BBox]:
+        """Helper method for parsing bounding box grid. It will try to parse it into `List[BBox]`"""
 
         if isinstance(bbox_grid, list):
-            return BBoxCollection(bbox_grid)
+            return bbox_grid
 
-        raise ValueError(f"Parameter 'bbox_grid' should be an instance of {BBoxCollection}")
+        raise ValueError("Parameter 'bbox_grid' should be an instance of List[BBox]")
```
zigaLuksic, line 436: this method is now useless.

sentinelhub/areas.py
```diff
@@ -407,7 +407,7 @@ def __init__(
         self,
         shape_list: Iterable[Union[Polygon, MultiPolygon, _BaseGeometry]],
         crs: CRS,
-        bbox_grid: Union[List[BBox], BBoxCollection],
+        bbox_grid: List[BBox],
```
zigaLuksic, line 410: this can and should be upgraded to `Iterable[BBox]`

sentinelhub/areas.py
```diff
@@ -427,15 +427,13 @@ def __init__(
         super().__init__(shape_list, crs, **kwargs)
```
zigaLuksic, line 427: Github is being annoying and won't let me mark L425, `self.bbox_grid = self._parse_bbox_grid(bbox_grid)`

It is now clear that in that line we just assign input to an attribute, which is very dangerous if we are given mutable input structures. This was indirectly done before, but wasn't clear due to the `BBoxCollection` wrapper (though i described it's issues in the internal report).

Change to `list(bbox_grid)` or similar.  Check that this way its also safe to use `BBoxCollection` as input, so that we're sure we're not breaking any code (yet)

sentinelhub/geometry.py
```diff
@@ -531,105 +531,3 @@ def _parse_geometry(geometry: Union[Polygon, MultiPolygon, dict, str]) -> Union[
             raise ValueError(f"Supported geometry types are polygon and multipolygon, got {type(geometry)}")
 
         return geometry
-
-
-class BBoxCollection(_BaseGeometry):
```
zigaLuksic, line 536: only deprecate for now. If we decide to do aggressive removal in the next version, we'll throw it out then. Until then I want the option to do a minor release without breaking code.

sentinelhub/geometry.py
```diff
@@ -531,3 +532,101 @@ def _parse_geometry(geometry: Union[Polygon, MultiPolygon, dict, str]) -> Union[
             raise ValueError(f"Supported geometry types are polygon and multipolygon, got {type(geometry)}")
 
         return geometry
+
+
+@deprecated_class(message_suffix="You could use list of BBox in in its place.")
```
zigaLuksic, line 537:
```suggestion
@deprecated_class(message_suffix="Use sequences of BBox objects instead.")
```

sentinelhub/areas.py
jgersak: Removed.

sentinelhub/areas.py
jgersak: Tested for BBoxCollection.

sentinelhub/geometry.py
jgersak: Done.

PR #393, opened 2023-02-07
Cleanup of geometry tests

zigaLuksic:


tests/test_geometry.py
```diff
@@ -1,190 +1,183 @@
...
 
 
 def test_bbox_to_str() -> None:
-    x1, y1, x2, y2 = 45.0, 12.0, 47.0, 14.0
-    crs = CRS.WGS84
-    expect_str = f"{x1},{y1},{x2},{y2}"
-    bbox = BBox(((x1, y1), (x2, y2)), crs)
-    assert str(bbox) == expect_str
+    bbox = BBox(((45.0, 12.0, 47.0, 14.0)), CRS.WGS84)
+    assert str(bbox) == "45.0,12.0,47.0,14.0"
```
jgersak, line 55: add test for `reverse` and not sorted input

zigaLuksic: the `__str__` method is probably gonna get :fire: after this MR

not sure we should test unsorted input as it is kind of an 'implementation detail'. We never mention that we're doing it. I'm not sure it's even correct of us to sort input, perhaps a warning would be better.

jgersak: No need to test ash ;)

tests/test_geometry.py
```diff
@@ -1,190 +1,183 @@
...
+
+def test_bbox_geometry_attribute() -> None:
+    bbox = BBox((0, 0, 1, 1), CRS.WGS84)
     assert isinstance(bbox.geometry, shapely.geometry.Polygon)
+    assert bbox.geometry.equals(shapely.geometry.Polygon([[0, 0], [0, 1], [1, 1], [1, 0], [0, 0]]))
 
 
-def test_buffer() -> None:
-    bbox = BBox([46.07, 13.23, 46.24, 13.57], CRS.WGS84)
+def test_bbox_buffer() -> None:
```
jgersak, line 129: Could this test case be parameterized like (test bbox, relative, expected ) 
I would also like to add simple bbox (0, 0, 10, 10) or sth.

zigaLuksic: the test cases are quite different for 'relative' and 'not relative'.

The BBox is intentionally a bit ugly so that it doesn't only work on pretty cases :) But the calculations are simple enough here that one can doublecheck they are correct.

zigaLuksic: we could parametrize the test BBox to have one pretty and one ugly :thinking:

jgersak (approved):

PR #397, opened 2023-02-09
Simplify io utils

zigaLuksic:
There was no reason to have all those miniature functions, since they weren't used anywhere. Some not even in the magic functions :joy:


jgersak (approved):
L(Perfect)TM

PR #401, opened 2023-02-13
Improve test data collections

jgersak:


tests/test_data_collections.py
```diff
@@ -26,21 +18,39 @@ def test_derive() -> None:
...
+)
+def test_collection_repr(definition_input: Dict[str, Any], expected: str) -> None:
+    assert repr(DataCollectionDefinition(**definition_input)) == expected
+
+
+def test_collection_equality() -> None:
     def1 = DataCollectionDefinition(api_id="X", _name="A")
     def2 = DataCollectionDefinition(api_id="X", _name="B")
 
     assert def1 == def2
```
zigaLuksic, line 43: are they the same object? (check with `is`)
how about definitions that should not be the same?
how about definitions that are the same but are not just one-parameter long?
Are these tests for collections or collection definitions?

tests/test_data_collections.py
```diff
@@ -26,21 +18,39 @@ def test_derive() -> None:
...
-        data_collection = DataCollection.define(
-            "NEW", api_id="X", sensor_type="Sensor", bands=("B01",), is_timeless=True
-        )
+    data_collection = DataCollection.define("NEW", api_id="X", sensor_type="Sensor", bands=("B01",), is_timeless=True)
 
     assert data_collection == DataCollection.NEW
+    assert DataCollection.NEW.api_id == "X"
 
+
+def test_define_faild() -> None:
```
zigaLuksic, line 53:
```suggestion
def test_define_fails() -> None:
```

zigaLuksic: add docstring/comment why this must fail. If you can, perhaps even make the name state the reason for failure, such as `test_define_same_api_id_fails`

Does this fail only because `test_define` gets run before it?

tests/test_data_collections.py
```diff
@@ -50,36 +60,47 @@ def test_define() -> None:
...
-        assert data_collection.collection_id == byoc_id
+    assert DataCollection.MY_BATCH.is_batch
+    assert not DataCollection.SENTINEL2_L2A.is_batch
 
 
-def test_attributes() -> None:
-    data_collection = DataCollection.SENTINEL3_OLCI
+@pytest.mark.parametrize("data_collection", [DataCollection.SENTINEL3_OLCI, DataCollection.SENTINEL2_L2A])
+@pytest.mark.parametrize("attribut", ["api_id", "catalog_id", "wfs_id", "service_url", "bands", "sensor_type"])
+def test_attributes(data_collection: DataCollection, attribut: str) -> None:
```
zigaLuksic, line 97: correct `attribut` to `attribute`

tests/test_data_collections.py
```diff
@@ -89,25 +110,37 @@ def test_attributes() -> None:
...
-    assert DataCollection.SENTINEL2_L2A.contains_orbit_direction("descending")
+@pytest.mark.parametrize(
+    "test_collection, expaected",
+    [
+        (DataCollection.SENTINEL2_L1C, False),
+        (DataCollection.SENTINEL1_EW, True),
+        (DataCollection.LANDSAT_TM_L1, False),
+    ],
+)
+def test_is_sentinel1(test_collection: DataCollection, expaected: bool) -> None:
```
zigaLuksic, line 121: correct `expaected` (also in following tests)

tests/test_data_collections.py
```diff
@@ -89,25 +110,37 @@ def test_attributes() -> None:
...
 def test_get_available_collections() -> None:
     collections = DataCollection.get_available_collections()
-    assert helper_check_collection_list(collections)
-
-
-def helper_check_collection_list(collection_list: Any) -> bool:
-    is_list = isinstance(collection_list, list)
-    contains_collections = all(isinstance(data_collection, DataCollection) for data_collection in collection_list)
-    return is_list and contains_collections
+    assert len(collections) >= 28
```
zigaLuksic, line 142: why is it 28? if we define a new one, will it be 29? if the user defines a new one, will it be then 30? do byocs count? do batch collections?

tests/test_data_collections.py
```diff
@@ -89,25 +110,37 @@ def test_attributes() -> None:
...
+    assert all([type(collection) == DataCollection for collection in collections])
```
zigaLuksic, line 143: don't compare types, use `isinstance`

tests/test_data_collections.py
jgersak: they should not be the same object just equal, more cases added

tests/test_data_collections.py
jgersak: done

tests/test_data_collections.py
jgersak: done

tests/test_data_collections.py
jgersak: Removed ... 28 are set on init, answer yes, yes, yes and yes.

tests/test_data_collections.py
jgersak: Done.

tests/test_data_collections.py
zigaLuksic: add comments for reasons why it should fail

tests/test_data_collections.py
zigaLuksic: would the tests look nicer if they were split into two? (one for equal, one for not-equal)

tests/test_data_collections.py
zigaLuksic: but how do you know that is the case? Maybe you should test it :P

tests/test_data_collections.py
jgersak: Yes, done.

tests/test_data_collections.py
jgersak: Done.

tests/test_data_collections.py
jgersak: Tested.

tests/test_data_collections.py
```diff
@@ -9,13 +9,25 @@
...
+)
+def test_derive(
+    data_colection_def: DataCollectionDefinition, derive_attributes: Dict[str, Any], expected_attributes: Dict[str, Any]
+) -> None:
+    derived_definition = data_colection_def.derive(**derive_attributes)
 
-    assert derived_definition.api_id == "X"
-    assert derived_definition.wfs_id == "Z"
-    assert derived_definition.collection_type is None
+    for atribure, value in expected_attributes.items():
```
zigaLuksic, line 29: `attribute`

PR #405, opened 2023-02-15
Fix splitter issue when batch geometry is missing

### LINKED ISSUE #337: [BUG] BatchSplitter doesn't work with bbox, only with geometry
**Describe the bug**

The above happens if a batch request is defined with the `bbox` parameter and not the `geometry` parameter. This causes issues with the following code.

```python
from sentinelhub import BatchSplitter
from matplotlib.patches import Patch

splitter = BatchSplitter(batch_request=batch_request, config=config)
splitter.get_bbox_list()
```

the error

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/var/folders/f7/wf5f4ngj0mxd5xrdjzqz7vyr0000gn/T/ipykernel_57279/4253228497.py in <module>
      2 from matplotlib.patches import Patch
      3 
----> 4 splitter = BatchSplitter(batch_request=batch_request, config=config)
      5 splitter.get_bbox_list()

~/.pyenv/versions/3.8.7/envs/am/lib/python3.8/site-packages/sentinelhub/areas.py in __init__(self, request_id, batch_request, config)
    714         self.batch_request = batch_request
    715 
--> 716         batch_geometry = batch_request.geometry
    717         super().__init__([batch_geometry.geometry], batch_geometry.crs)
    718 

~/.pyenv/versions/3.8.7/envs/am/lib/python3.8/site-packages/sentinelhub/sentinelhub_batch.py in geometry(self)
    648         _, geometry, crs = self._parse_bounds_payload()
    649         if geometry is None:
--> 650             raise ValueError('Geometry is not defined for this batch request')
    651         return Geometry(geometry, crs)
    652 

ValueError: Geometry is not defined for this batch request
```

The issue is in the assumption made in https://github.com/sentinel-hub/sentinelhub-py/blob/develop/sentinelhub/areas.py#L647

**To Reproduce**

1. define request with `bbox`
2. define batch request
3. use batch splitter

**Expected behavior**

no crash

**Environment**

- Python 3.8.7
- latest `develop` of `sentinelhub-py`

**Stack trace or screenshots**

image

**Desktop (please complete the following information):**

- macOS Monterey 12.6
 - VSCode

zigaLuksic:
This is one of those "Its so easy, and yet so hard." problems.

The idea is really simple, if there is no `geometry`, just take the geometry of the `bbox`. But since both getters are throwing errors upon non-definition, we only have a private method that lets us check. But that again does not seem right, since it's a private method...

I am open to suggestions.

Closes #337

### DIFF
--- a/sentinelhub/api/batch/process.py
+++ b/sentinelhub/api/batch/process.py
@@ -533,28 +533,24 @@ class BatchRequest(BaseBatchRequest):  # pylint: disable=abstract-method
         return self.process_request["evalscript"]
 
     @property
-    def bbox(self) -> BBox:
+    def bbox(self) -> Optional[BBox]:
         """Provides a bounding box used by a batch request
 
         :return: An area bounding box together with CRS
         :raises: ValueError
         """
         bbox, _, crs = self._parse_bounds_payload()
-        if bbox is None:
-            raise ValueError("Bounding box is not defined for this batch request")
-        return BBox(bbox, crs)  # type: ignore[arg-type]
+        return None if bbox is None else BBox(bbox, crs)  # type: ignore[arg-type]
 
     @property
-    def geometry(self) -> Geometry:
+    def geometry(self) -> Optional[Geometry]:
         """Provides a geometry used by a batch request
 
         :return: An area geometry together with CRS
         :raises: ValueError
         """
         _, geometry, crs = self._parse_bounds_payload()
-        if geometry is None:
-            raise ValueError("Geometry is not defined for this batch request")
-        return Geometry(geometry, crs)
+        return None if geometry is None else Geometry(geometry, crs)
 
     def _parse_bounds_payload(self) -> Tuple[Optional[List[float]], Optional[list], CRS]:
         """Parses bbox, geometry and crs from batch request payload. If bbox or geometry don't exist it returns None
--- a/sentinelhub/areas.py
+++ b/sentinelhub/areas.py
@@ -653,8 +653,9 @@ class BatchSplitter(AreaSplitter):
         self.tile_size = self._get_tile_size()
         self.tile_buffer = self._get_tile_buffer()
 
-        _, geom_def, _ = batch_request._parse_bounds_payload()
-        batch_geometry: _BaseGeometry = batch_request.geometry if geom_def else batch_request.bbox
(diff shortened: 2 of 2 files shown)


mlubej:
I would need some context so that I can understand this properly. In what situations does the existing batch request not have a defined geometry? Is it after creating the batch request but before running the analysis?

zigaLuksic:
> I would need some context so that I can understand this properly. In what situations does the existing batch request not have a defined geometry? Is it after creating the batch request but before running the analysis?

I would guess it's in the cases where the user only provides `bbox` but no a `geometry`. In the SH BatchAPI docs it states (for `bbox`): "The request bounding box. This and/or GEOMETRY must be specified."

mlubej:
> I am open to suggestions.

How about expanding the `bbox`/`geometry` methods in `sentinelhub/api/batch/process.py` with a `raise: bool` parameter with a default to `True`, then you could use the public getter for the geometry and keep the private method intact and unused in this case.

> This and/or GEOMETRY must be specified.

In this case the above suggestion makes especially more sense, since raising the error if either one doesn't exist doesn't make much sense with this specification.

zigaLuksic:
> > I am open to suggestions.
> 
> How about expanding the `bbox`/`geometry` methods in `sentinelhub/api/batch/process.py` with a `raise: bool` parameter with a default to `True`, then you could use the public getter for the geometry and keep the private method intact and unused in this case.

Sadly properties and flags don't mix.
image

I could make two methods `get_bbox'` and `get_geometry` that return `Optional` values (and could be called by the properties). But that feels like boilerplate. It is again a problem of "the interface is public and therefore hard to adjust".

mlubej:
> Sadly properties and flags don't mix.

Then maybe the proper question is if this property should raise a `ValueError` in the first place. If the `bbox` or `geometry` don't exist and if there is a situation where this is completely fine, then the code should reflect this and I would suggest removing the raise statements here and simply return Nones. We have to check then where it was expected that this should have raised an error and fix it accordingly

mlubej:
So, in addition, perhaps it's the `_parse_bounds_payload` method that should raise an error if both `bbox` AND `geometry` are `None` (unless I misunderstood and this is also a valid case).

mlubej (approved):
Much better now, I think. Lucky that this doesn't break anything later on.. But in any case, it's definitely a move towards a better representation of what is.

PR #407, opened 2023-02-16
Fix/statistical request example key value error

urhprimozic:
Changed `column_name = f"bands_{band}_{stat}"` to `column_name = f"bands_{band.name}_{stat}"` to solve keyError.


mlubej:
@urhprimozic Can you please check if there is any other occurrences where bands from `DataCollection.<SOMETHING>.bands` are used and the `.name` attribute is missing?

zigaLuksic (approved):

PR #412, opened 2023-03-08
166 fix get failed statistical requests

chorng:


sentinelhub/data_utils.py
```diff
@@ -83,6 +84,18 @@ def _extract_response_data(response_data: List[JsonDict], exclude_stats: List[st
...
+    """Check statistical request type
+
+    :param request_data: An input representation of (Batch) Statistical API result
+        returned from `DataRequest.get_data()`.
+    :return: bool
+    """
+    if "id" in result_data[0]:
+        return True
+    return False
+
```
Adrok24, line 97: what about this way in one line?
    return "id" in result_data[0]

sentinelhub/data_utils.py
```diff
@@ -116,37 +129,59 @@ def statistical_to_dataframe(result_data: List[JsonDict], exclude_stats: Optiona
...
     :return: The identifier of a geometry that has a response status of PARTIAL and the failed intervals.
     """
-    failed_intervals = []
+    failed_intervals: List[Tuple[str, str]] = []
     for interval in response_data:
         if "error" in interval:
             failed_intervals.append((interval["interval"]["from"], interval["interval"]["to"]))
 
-    return {"identifier": identifier, "failed_intervals": failed_intervals} if failed_intervals else None
+    return {"Failed_intervals": failed_intervals} if failed_intervals else None
```
Adrok24, line 145: why "Failed_intervals with the first in capital letters?

sentinelhub/data_utils.py
```diff
@@ -116,37 +129,59 @@ def statistical_to_dataframe(result_data: List[JsonDict], exclude_stats: Optiona
...
-            failed_requests.append({"identifier": identifier})
+        failed_request = {"Identifier": result["identifier"]}
+        if "error" in result:
+            failed_intervals: Any = _FULL_TIME_RANGE_FAILED
         else:
-            failed_intervals = _get_failed_intervals(identifier, response["data"])
-            if failed_intervals:
-                failed_requests.append(failed_intervals)
+            response = result["response"]
+            if not response:
```
Adrok24, line 161: what about using the get function, like response = result.get('response')

sentinelhub/data_utils.py
```diff
@@ -116,37 +129,59 @@ def statistical_to_dataframe(result_data: List[JsonDict], exclude_stats: Optiona
...
     """
     failed_requests = []
     for result in result_data:
-        identifier, response = result["identifier"], result["response"]
-        if not response:
-            failed_requests.append({"identifier": identifier})
+        failed_request = {"Identifier": result["identifier"]}
+        if "error" in result:
+            failed_intervals: Any = _FULL_TIME_RANGE_FAILED
         else:
```
Adrok24, line 159: i would also rewrite this part too to avoid this failed_intervals: Any  also

Adrok24 (commented):
is getting much more tidy, i havent tested all. But it seems nicer.

sentinelhub/data_utils.py
```diff
@@ -130,23 +138,42 @@ def _get_failed_intervals(
...
+        if "error" in result or not result["response"]:
+            failed_requests.append({"identifier": identifier, "failed_intervals": "full time range"})
+            continue
+
+        failed_intervals = _get_failed_intervals(result["response"]["data"])
+        if failed_intervals:
+            failed_requests.append({"identifier": identifier, "failed_intervals": failed_intervals})
 
     return failed_requests
+
```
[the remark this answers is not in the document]
zigaLuksic: what if `_get_failed_intervals` instead accepted a full result` and returned `"full time range"` in case there is an error?
Then you could perhaps do something like: 
```python
failed_intervals = [{"identifier": identifier, "failed_intervals": _get_failed_intervals(result)} for result in resultdata]
return [failure_report for failure_report in failed_intervals if failure_report["failed_intervals"] is not None]
```

sentinelhub/data_utils.py
```diff
@@ -83,6 +83,16 @@ def _extract_response_data(response_data: List[JsonDict], exclude_stats: List[st
     return df_entries
 
 
+def _is_batch_stat(result_data: JsonDict) -> bool:
+    """Check statistical request type
+
+    :param request_data: An input representation of (Batch) Statistical API result
+        returned from `DataRequest.get_data()`.
+    :return: bool
+    """
```
zigaLuksic, line 92:
```suggestion
    """Identifies whether the resulting data belongs to a batch statistical request or not."""
```

I suppose this is the idea of this function?

sentinelhub/data_utils.py
```diff
@@ -116,37 +126,46 @@ def statistical_to_dataframe(result_data: List[JsonDict], exclude_stats: Optiona
...
+    if _is_batch_stat(result_data):
+        return (
+            "full time range"
+            if "error" in result_data or not result_data["response"]
+            else [
+                (interval["interval"]["from"], interval["interval"]["to"])
+                for interval in result_data["response"]["data"]
+                if "error" in interval
+            ]
+        )
```
zigaLuksic, line 144: this kind of nesting is very hard to read in python. Move the conditional out of the `return` statement. 

You also use `"full time range"` in many places, so you should extract it as a constant

sentinelhub/data_utils.py
```diff
@@ -116,37 +126,46 @@ def statistical_to_dataframe(result_data: List[JsonDict], exclude_stats: Optiona
...
-            if failed_intervals:
-                failed_requests.append(failed_intervals)
-
-    return failed_requests
+    if _is_batch_stat(result_data[0]):
+        return [
+            {"identifier": result["identifier"], "failed_intervals": _get_failed_intervals(result)}
+            for result in result_data
+            if _get_failed_intervals(result)
+        ]
```
zigaLuksic, line 168: This doesn't seem quite right. `_get_failed_intervals` returns a string or a tuple, so why are you doing `if _get_failed_intervals(result)`?

sentinelhub/data_utils.py
```diff
@@ -116,37 +126,46 @@ def statistical_to_dataframe(result_data: List[JsonDict], exclude_stats: Optiona
...
-        identifier, response = result["identifier"], result["response"]
-        if not response:
-            failed_requests.append({"identifier": identifier})
-        else:
-            failed_intervals = _get_failed_intervals(identifier, response["data"])
-            if failed_intervals:
-                failed_requests.append(failed_intervals)
-
-    return failed_requests
+    if _is_batch_stat(result_data[0]):
```
zigaLuksic, line 163: Here you split on whether it is a batch response or a regular one. But in `_get_failed_intervals` you do the same. I suggest you only do it here, and unify `_get_failed_intervals` for both. If you can't, then split it into `_get_failed_intervals_batch` and `_get_failed_intervals_regular` (but i am 95% sure you can do all the 'special' handling here)

sentinelhub/data_utils.py
chorng: It returns a string, a list of tuples, or an empty list if there's no error in any intervals.

zigaLuksic: ah, right. Make it more explicit with `if _get_failed_intervals(result) != []` or something. We have had many issues due to not being specific with `if x` instead of `if x is None` and similar.

zigaLuksic: I am also not a fan of very "busy" list comprehensions, and prefer doing them in steps. How does this sound to you:
```python
failed_intervals = ((result["identifier"], _get_failed_intervals(result)) for result in result_data)
    return [
        {"identifier": identifier, "failed_intervals": intervals}
        for identifier, intervals in failed_intervals
        if intervals != []
    ]
```

chorng: I'm actually confused when to/ not to use list comprehensions atm. I feel it is a bit conflicting with avoid `if else`, `continue`, `break`, etc. Is there a general guild line for this?

zigaLuksic: If you have a more difficult control flow (e.g. multiple `if`s or `for` loops) in a list comprehension, which spans multiple lines, it should be refactored. That does not mean not to use list comprehension, it just means you need more steps.

When in doubt, i suggest you write out both options and choose the one that is more readable. Readability > anything else (unless performance is a priority).

In this case we have
```python
failed_intervals = []
for result in result_data:
    intervals_for_result = _get_failed_intervals(result)
    if intervals_for_result != []:
        failed_intervals.append({"identifier": result["identifier"], "failed_intervals": intervals_for_result})

return failed_intervals
```
versus

```python
        return [
            {"identifier": result["identifier"], "failed_intervals": _get_failed_intervals(result)}
            for result in result_data
            if _get_failed_intervals(result)
        ]
```

You can see that the first one has more lines of code, requires more variable names. And it tells you *how* to build the result.
The second one tells you more about *what* is being returned. In general that is a bit favored for readability. But IMO large list comprehensions suffer from the same problem, where we just effectively move all the control flow inside a list comprehension. In that case it is probably better to do it in multiple steps.

```python
    failed_intervals = ((result["identifier"], _get_failed_intervals(result)) for result in result_data)
    return [
        {"identifier": ident, "failed_intervals": intervals} for ident, intervals in failed_intervals if intervals != []
    ]
```

Here I tried to separate two steps. The first one calculates failed intervals for all the results. The second one filters and repackages them.

But in the end a lot of this is subjective.

sentinelhub/data_utils.py
chorng: I'm struggling a bit here. A failed response could be:
- {"id": 1, "identifier": "xxx", "error": {}}
- {"id": 1, "identifier": "xxx", "response": null}
- {"id": 1, "identifier": "xxx", "response": {"data": ["interval": {}, "error": {}]}}
- {"data": ["interval": {}, "error": {}]}
- {"error": {}}

May I have your suggestions on how to build the function for these conditions?

zigaLuksic: so you need a `_get_failed_intervals_response` and then `_get_failed_intervals_batch` can check if it got a response, and if it did it uses `_get_failed_intervals_response`, that should work right?

chorng: Ah true, will fix this.

sentinelhub/data_utils.py
zigaLuksic: i see you changed the first line, but i would remove the docstring about parameters as well
the typess say all that one needs, especially since the `request_data` description seems odd

sentinelhub/data_utils.py
```diff
@@ -161,11 +156,12 @@ def get_failed_statistical_requests(result_data: List[JsonDict]) -> List[JsonDic
...
-            for result in result_data
-            if _get_failed_intervals(result)
+            {"identifier": ident, "failed_intervals": intervals}
+            for ident, intervals in batch_failed_response
+            if intervals != []
         ]
 
-    failed_intervals = _get_failed_intervals(result_data[0])
-    return [{"failed_intervals": failed_intervals}] if failed_intervals else []
+    failed_intervals = _FULL_TIME_RANGE if "error" in result_data[0] else _get_failed_intervals(result_data[0]["data"])
```
zigaLuksic, line 166: why do we always look at the zero element here? why are we sure there is no others?

chorng: If it is a normal statistical request there will be only one result in the list

zigaLuksic: but it would be almost no extra work to make it work on any list (and we then don't need to worry about the implicit assumption that there will always be just 1)

chorng: I make it go through the list and use a defult index as the identifier for normal stat

sentinelhub/data_utils.py
```diff
@@ -116,37 +122,47 @@ def statistical_to_dataframe(result_data: List[JsonDict], exclude_stats: Optiona
...
+        ]
+
+    normal_failed_intervals = (
+        _FULL_TIME_RANGE if "error" in result else _get_failed_intervals(result["data"]) for result in result_data
+    )
+    return [
+        {"identifier": index, "failed_intervals": intervals}
+        for index, intervals in enumerate(normal_failed_intervals)
+        if intervals != []
+    ]
```
zigaLuksic, line 168: Excellent, now we can try and do this!
```suggestion
     if _is_batch_stat(result_data[0]):
        failed_responses = ((result["identifier"], _get_failed_batch_response(result)) for result in result_data)
    else:
        failed_responses = enumerate(
            _FULL_TIME_RANGE if "error" in result else _get_failed_intervals(result["data"]) for result in result_data
        )

    return [
        {"identifier": idx, "failed_intervals": intervals} for idx, intervals in failed_responses if intervals != []
    ]
```

Doublecheck this actually works :see_no_evil:

chorng: It works but in this case mypy doesn't like the assignment of generator and enumerate to the same variable, which may lead to something looks bad like `failed_responses: Any = ...`

Is there a preferable way to solve this?

chorng: Otherwise I see that `if else` in this case make the script much shorter and cleaner :)

zigaLuksic: I would do this:
```python
    failed_responses: Iterable[Tuple[Any, Union[str, List[Tuple[str, str]]]]]

    if _is_batch_stat(result_data[0]):
        failed_responses = ((result["identifier"], _get_failed_batch_response(result)) for result in result_data)
    else:
        failed_responses = enumerate(
            _FULL_TIME_RANGE if "error" in result else _get_failed_intervals(result["data"]) for result in result_data
        )
    return [
        {"identifier": idx, "failed_intervals": intervals} for idx, intervals in failed_responses if intervals != []
    ]
```

PR #414, opened 2023-03-15
Fix failing docs and update docs requirements

mlubej:
Changes:
- replaces `m2r2` with `sphinx_mbinclude`
- removes `m2r2` and `jinja2`
- update sphinx errors/issues in docs
- update sphinx error with nbconvert in notebook

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -22,7 +22,7 @@ from typing import Any, Dict, Optional
 
 # General information about the project.
 project = "Sentinel Hub"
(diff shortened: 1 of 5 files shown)


mlubej:
There are still some sphinx errors when building the html, but I think they are safe to ignore:

image

- the first one is probably an issue due to `INTRO.md` being created inplace
- the second one is due to a deprecated module

zigaLuksic (approved):

PR #420, opened 2023-03-16
Update test_geo_utils

jgersak:
parametrize existing tests and add test cases
add test for pixel_to_utm, utm_to_pixel, wgs84_to_pixel

### DIFF
--- a/tests/test_geo_utils.py
+++ b/tests/test_geo_utils.py
@@ -5,58 +5,70 @@ from typing import Tuple, Union
 
 import pytest
 
-from sentinelhub import CRS, BBox, geo_utils
+from sentinelhub import CRS, BBox
+from sentinelhub.geo_utils import (
+    bbox_to_dimensions,
+    bbox_to_resolution,
+    get_image_dimension,
+    get_utm_crs,
+    pixel_to_utm,
+    to_wgs84,
+    transform_point,
+    utm_to_pixel,
+    wgs84_to_pixel,
+    wgs84_to_utm,
+)
+
+BBOX_WGS84 = BBox(((111.644, 8.655), (111.7, 8.688)), CRS.WGS84)
+BBOX_UTM = BBox(((570851.8316965176, 956770.2133183606), (577006.4223865959, 960429.6742984429)), CRS("32649"))
+BBOX_POP_WEB = BBox(((12428153.230124235, 967155.4076988235), (12434387.12160866, 970871.4287682716)), CRS.POP_WEB)
 
 
-def test_wgs84_to_utm() -> None:
-    x, y = geo_utils.wgs84_to_utm(15.525078, 44.1440478, CRS.UTM_33N)
-    assert (x, y) == pytest.approx((541995.694062, 4888006.132887), rel=1e-8)
+@pytest.mark.parametrize("input_bbox, expected_bbox", [(BBOX_WGS84, BBOX_UTM), (BBOX_WGS84, BBOX_POP_WEB)])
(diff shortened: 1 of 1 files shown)


tests/test_geo_utils.py
```diff
@@ -5,58 +5,70 @@
...
 
+BBOX_WGS84 = BBox(((111.644, 8.655), (111.7, 8.688)), CRS.WGS84)
+BBOX_UTM = BBox(((570851.8316965176, 956770.2133183606), (577006.4223865959, 960429.6742984429)), CRS("32649"))
+BBOX_POP_WEB = BBox(((12428153.230124235, 967155.4076988235), (12434387.12160866, 970871.4287682716)), CRS.POP_WEB)
 
-def test_wgs84_to_utm() -> None:
-    x, y = geo_utils.wgs84_to_utm(15.525078, 44.1440478, CRS.UTM_33N)
-    assert (x, y) == pytest.approx((541995.694062, 4888006.132887), rel=1e-8)
 
+@pytest.mark.parametrize("input_bbox, expected_bbox", [(BBOX_WGS84, BBOX_UTM), (BBOX_WGS84, BBOX_POP_WEB)])
```
zigaLuksic, line 27: hold on, the function is named `wgs84_to_utm`, why does it work with `POP_WEB`?

tests/test_geo_utils.py
```diff
@@ -5,58 +5,70 @@
...
+    to_wgs84,
+    transform_point,
+    utm_to_pixel,
+    wgs84_to_pixel,
+    wgs84_to_utm,
+)
 
+BBOX_WGS84 = BBox(((111.644, 8.655), (111.7, 8.688)), CRS.WGS84)
+BBOX_UTM = BBox(((570851.8316965176, 956770.2133183606), (577006.4223865959, 960429.6742984429)), CRS("32649"))
+BBOX_POP_WEB = BBox(((12428153.230124235, 967155.4076988235), (12434387.12160866, 970871.4287682716)), CRS.POP_WEB)
```
zigaLuksic, line 24: why do these have an absurd amount of decimal points?
For the UTM BBox you're specifying nanometers!

tests/test_geo_utils.py
```diff
@@ -5,58 +5,70 @@
...
-    resx, resy = geo_utils.bbox_to_resolution(bbox, 512, 512)
+def test_get_utm_crs() -> None:
+    assert get_utm_crs(*BBOX_WGS84.lower_left) is BBOX_UTM.crs
+    assert get_utm_crs(*BBOX_WGS84.upper_right) is BBOX_UTM.crs
+
 
-    assert (resx, resy) == pytest.approx((12.0207, 7.1474), rel=1e-4)
+@pytest.mark.parametrize("input_bbox", [BBOX_WGS84, BBOX_UTM, BBOX_POP_WEB])
+def test_bbox_to_resolution(input_bbox: BBox) -> None:
+    assert bbox_to_resolution(input_bbox, 512, 512) == pytest.approx((12.0207, 7.1474), rel=1e-4)
```
zigaLuksic, line 49: you're still just testing just a single bbox right? just in different CRS

wouldn't it be better to have three different bboxes to test all this?

tests/test_geo_utils.py
zigaLuksic: if you're going to use the same bbox just reprojected a few times, then you can't really avoid having lots of decimals. But i think that tests would work better with three independant bboxes (outside of the tests for reprojection).

tests/test_geo_utils.py
jgersak: function have wrong name, test updated

zigaLuksic (approved):

PR #421, opened 2023-03-16
Add ingestion_start parameter to the BYOC tile dataclass

mlubej:
Changes:
- added the parameter to the dataclass
- minor extension to the tests so that the parameter is at least used
- updated a notebook where the metadata was updated and other minor modifications


zigaLuksic (approved):
thanks for rerunning the notebook <3

PR #424, opened 2023-03-17
Batch Statistical STOP status and action

jgersak:
deprecat cancel_job and add stop_job
add BatchRequestStatus and BatchUserAction
update test

### DIFF
--- a/sentinelhub/api/batch/statistical.py
+++ b/sentinelhub/api/batch/statistical.py
@@ -12,6 +12,7 @@ from dataclasses_json import config as dataclass_config
 from dataclasses_json import dataclass_json
 from typing_extensions import NotRequired
 
+from ...exceptions import deprecated_function
 from ...types import Json, JsonDict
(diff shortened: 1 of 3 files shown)


sentinelhub/api/batch/statistical.py
```diff
@@ -179,7 +181,18 @@ def cancel_job(self, batch_request: BatchStatisticalRequestType) -> Json:
         :param batch_request: It could be a batch request object, a raw batch request payload or only a batch
             request ID.
         """
-        return self._call_job(batch_request, "cancel")
+        return self._call_job(batch_request, "stop")
```
zigaLuksic, line 184: I know I suggested rerouting, but tbh since we have a deprecation warning, it might as well stay as it is :thinking: 
I'm reverting it as a suggestion so you don't have to do anything.
```suggestion
        return self._call_job(batch_request, "cancel")
```

zigaLuksic (approved):

PR #425, opened 2023-03-20
Further changes to SHConfig

zigaLuksic:
- removes dictionary-style access
- renames `get_config_dictionary` to the shorter and more common `to_dict` (am willing to accept new names as well). `hide_credentials` is now a parameter, the on at `__init__` only controls masking in `repr` and `str`
- class attributes specifying which attributes are regular fields and which are credentials have been changed to tuples. The `get_params` accessor is used instead of direct access, allowing us to remove duplication.

### DIFF
--- a/sentinelhub/config.py
+++ b/sentinelhub/config.py
@@ -10,6 +10,8 @@ import os
 from pathlib import Path
 from typing import Dict, Iterable, Optional, Tuple, Union
 
+from .exceptions import deprecated_function
+
 
 class SHConfig:  # pylint: disable=too-many-instance-attributes
     """A sentinelhub-py package configuration class.
@@ -78,8 +80,8 @@ class SHConfig:  # pylint: disable=too-many-instance-attributes
 
     def __init__(self, hide_credentials: bool = True, use_defaults: bool = False):
         """
-        :param hide_credentials: If `True` then methods that provide the entire content of the config object will mask
-            out all credentials. But credentials could still be accessed directly from config object attributes.
+        :param hide_credentials: If `True` then credentials will be masked when transforming the dictionary to a string.
+            Credentials can still be accessed directly from config object attributes.
         :param use_defaults: Does not load the configuration file, returns config object with defaults only.
         """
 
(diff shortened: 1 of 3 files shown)


sentinelhub/config.py
```diff
@@ -233,39 +223,45 @@ def _reset_param(self, param: str, default_config: SHConfig) -> None:
...
         """Get a dictionary representation of `SHConfig` class. If `hide_credentials` is set to `True` then
         credentials will be masked.
 
         :return: A dictionary with configuration parameters
         """
-        config_params = {param: getattr(self, param) for param in self.CONFIG_PARAMS}
+        return self.to_dict(hide_credentials=self._hide_credentials)
+
+    def to_dict(self, hide_credentials: bool = True) -> Dict[str, Union[str, float]]:
+        """Get a dictionary representation of `SHConfig` class.
```
mlubej, line 241:
```suggestion
        """Get a dictionary representation of the `SHConfig` class.
```

sentinelhub/config.py
```diff
@@ -233,39 +223,45 @@ def _reset_param(self, param: str, default_config: SHConfig) -> None:
...
     def get_config_dict(self) -> Dict[str, Union[str, float]]:
         """Get a dictionary representation of `SHConfig` class. If `hide_credentials` is set to `True` then
         credentials will be masked.
 
         :return: A dictionary with configuration parameters
         """
-        config_params = {param: getattr(self, param) for param in self.CONFIG_PARAMS}
+        return self.to_dict(hide_credentials=self._hide_credentials)
+
+    def to_dict(self, hide_credentials: bool = True) -> Dict[str, Union[str, float]]:
```
mlubej, line 240: I thought that in the previous version of the code it was meant that you could not print/see hidden fields if the config was initialized with credential masking set to true. Now we have a method where the masking can be controlled, which makes me wonder about its purpose.

Perhaps I'm missing a wider understanding of the masking here. IIUC, no one should be able to see the credentials' values if the config was initialized with masking, right? And this probably also goes for accessing values with `getattr`?

zigaLuksic: even before `getattr` always gave you the values no matter what you set for masking
It just controlled the behavior of `get_config_dict`. But because of this, the method was never used outside of `__repr__` and `__str__`.

By making the parameter explicit, it has other uses, and the init parameter only controls `__repr__` and `__str__`. If it were up to me I'd remove the `__init__` parameter and make the credentials always masked in the string representations.

Without being able to control if the dict is masked or not outside of the `__init__` this method is nearly useless, so if we go that route I'd make it private.

mlubej: So the question is if only `__repr__` and `__str__` are covered and `getattr` (and perhaps some other approaches) are not, does it make sense to have this half masking? Is it meant to prevent someone getting sensitive info, or is it more about helping out potentially clumsy users posting them on the web?

I imagine it's more about the latter, in which case I agree with you that we could go in the direction of removing the parameter from `__init__` and having it always set.  Regarding then removing the `to_dict`  or not, I don't have a strong opinion. I feel it's nice to have a convenience function.

Do you find these changes too invasive, or do you want to take them into account now that your ideas have extra support?

zigaLuksic: if `getattr` would be masked, the object itself becomes kinda useless (since you have credentials but cant access them).
Originally it was just a way for us to mask credentials in logs.

But we've since put more weight on security, so I believe that at this point `__repr__` and `__str__` could have masking hardcoded. The only reason not to have that is if the user wants to check that credentials have been set correctly in a notebook field. But then an output cell contains the sensitive data and it's all too easy to commit it to history.

PR #426, opened 2023-03-21
Add profiles to Config

zigaLuksic:


sentinelhub/config.py
```diff
@@ -144,44 +149,56 @@ def __eq__(self, other: object) -> bool:
         return all(getattr(self, param) == getattr(other, param) for param in self.get_params())
 
     @classmethod
-    def load(cls, filename: Optional[str] = None) -> SHConfig:
+    def load(cls, filename: Optional[str] = None, profile: str = DEFAULT_PROFILE) -> SHConfig:
```
mlubej, line 152: filenames now don't make sense anymore, right? do you plan this for a separate MR?

sentinelhub/config.py
```diff
@@ -10,11 +10,16 @@
...
+import tomli_w
+
+DEFAULT_PROFILE = "default"
+
 
 class SHConfig:  # pylint: disable=too-many-instance-attributes
     """A sentinelhub-py package configuration class.
 
-    The class reads the configurable settings from ``config.json`` file on initialization:
+    The class reads the configurable settings from ``config.toml`` file on initialization:
```
mlubej, line 22: There are still a bunch of instances of mentions of `config.json` in the repo, did you plane to change these parts as well?

mlubej (commented):
Looks good so far, aside from the comments.

One extra thing is the fact that the full profile is saved, while I would prefer to have only the updated fields in the config, so that the config file is lighter, and the defaults are used for the non-specified one. 

Could be you're planning for this at a later step, but wanted to point it out just in case.

sentinelhub/config.py
zigaLuksic: there is an open issue for adjusting docs, want to wait until we solidify everything

sentinelhub/config.py
zigaLuksic: in some cases one might want to have a file in a different location i guess?
for instance in the SH-QGIS plugin etc. But yeah, it does complicate things a bit in the loading method.

zigaLuksic:
> Looks good so far, aside from the comments.
> 
> One extra thing is the fact that the full profile is saved, while I would prefer to have only the updated fields in the config, so that the config file is lighter, and the defaults are used for the non-specified one.
> 
> Could be you're planning for this at a later step, but wanted to point it out just in case.

yes, an issue has already been opened for this.

sentinelhub/config.py
zigaLuksic: I guess it just feels weird to not have the standard `object.save(file)` and `load_object(file)` interface :thinking:  will sleep it through

zigaLuksic: i'll leave it in for now, and open an issue so that we can reconsider this decision

PR #428, opened 2023-03-22
Refactor config tests

zigaLuksic:
1. Did some refactoring (mostly renaming)
2. Removed old `ray` test because it was added due to issues with the global cache, which we have since removed

I mostly glanced it over, so any further suggestions are welcome.


jgersak (approved):
LGTM

PR #429, opened 2023-03-22
Minimize what is saved in config

zigaLuksic:
the default user is also renamed to  `default-user` to prevent users from thinking it represents default values

### DIFF
--- a/sentinelhub/config.py
+++ b/sentinelhub/config.py
@@ -13,7 +13,7 @@ from typing import Dict, Iterable, Optional, Tuple, Union
 import tomli
 import tomli_w
 
-DEFAULT_PROFILE = "default"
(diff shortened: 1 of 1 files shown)


zigaLuksic:
I tested and saving all-defaults results in a file with contents:
```toml
[default-profile]

```

But that is still correctly loaded, and if you save another profile you get
```toml
[default-profile]

[beep]
instance_id = "no-drama-llama"
```
and loading either profiles still works as expected

jgersak (approved):
Looks good, but adding test would not be bad :)

zigaLuksic:
> Looks good, but adding test would not be bad :)

really difficult to automate this one i'm afraid. as long as the save/load tests go through, it should be fine enough.

PR #437, opened 2023-03-28
Adapt Config to be a dataclass

zigaLuksic:
Also allows us switching to pydantic in the future :dagger: 

The only thing we need to fully agree on is wether `init params > env > config.toml` or `env > init params > config.toml`.
Previously `profile` was set so that env trumps all, which is why the rest of the params follow suit.

### DIFF
--- a/sentinelhub/config.py
+++ b/sentinelhub/config.py
@@ -49,7 +49,7 @@ class _SHConfig:
             raise ValueError("Value of config parameter `max_opensearch_records_per_query` must be at most 500")
 
 
-class SHConfig(_SHConfig):  # pylint: disable=too-many-instance-attributes
+class SHConfig(_SHConfig):
     """A sentinelhub-py package configuration class.
 
     The class reads the configurable settings from ``config.toml`` file on initialization:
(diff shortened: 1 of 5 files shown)


sentinelhub/config.py
```diff
@@ -69,132 +92,61 @@ class SHConfig:  # pylint: disable=too-many-instance-attributes
...
-        "download_sleep_time",
-        "download_timeout_seconds",
-        "number_of_download_processes",
-    )
 
-    def __init__(self, profile: str = DEFAULT_PROFILE, *, use_defaults: bool = False):
+    def __init__(self, profile: str = DEFAULT_PROFILE, *, use_defaults: bool = False, **kwargs: Any):
         """
         :param profile: Specifies which profile to load form the configuration file. The environment variable
             SH_USER_PROFILE has precedence.
```
mlubej, line 99: I feel like the `$` makes it more intuitive

```suggestion
        :param profile: Specifies which profile to load from the configuration file. The environment variable
            $SH_USER_PROFILE has precedence.
```

zigaLuksic: why? :thinking:

mlubej: mah.. I looked at it again now and maybe not, it's even OS specific.

mlubej (approved):
Tested it out, seems to work as expected.

PR #440, opened 2023-03-29
pyproject.toml based install

zigaLuksic:
I know we're all used to the familiar `requirements.txt` format, but since we've been cleaning up installations and building, I thought to myself "why not switch to the currently suggested approach". Setting stuff in `setup.py` is discouraged unless really necessary, and `setup.cfg` is yet another file, so I skipped straight ahead into `pyproject.toml` based builds.

Since we no longer have a `setup.py` we need to build via the `build` package (it's ungooglable, horrible name choice). The backend is `hatchling` because `setuptools` don't let me exclude tests without a `MANIFEST.in`.


mlubej (commented):
Tried it in a fresh env and seems to work. I have to say I'm not a hater of the requirements file, but I do dislike it when there are multiple requirements files, so this is a nice solution to the problem.

tests work also locally with a fresh install!

zigaLuksic:
> Tried it in a fresh env and seems to work. I have to say I'm not a hater of the requirements file, but I do dislike it when there are multiple requirements files, so this is a nice solution to the problem.
> 
> tests work also locally with a fresh install!

I feel like the `requirements.txt` list the packages a bit more neatly (no quotes and indentations etc.)
But from what i gather, dependencies should be soft locked (lower bounds) while `requirements.txt` should have hard locks, and you are encouraged to have both. Dependencies make sure that installation goes through, but if some package wrecks your environement, one can install it with `requirements.txt` and be sure it will work. But i dont think we need that just yet

PR #441, opened 2023-03-29
Move evalscript generation to sh-py

mlubej:
Changes:
- move utility from https://github.com/sentinel-hub/eo-learn/blob/develop/io/eolearn/io/sentinelhub_process.py#L478 to this repo
- add tests

The evalscript generation is simple in nature, and offers the following:
- limited to `SIMPLE` mosaicking
- only download bands (separate or merged into a single output)
- use DN or not (if available)
- provide a subset of bands/meta_bands

### DIFF
--- a/sentinelhub/evalscript.py
+++ b/sentinelhub/evalscript.py
@@ -1,0 +1,117 @@
+"""
+Module defining evalscript generation utilities
+"""
+from typing import Dict, List, Optional
+
+import numpy as np
+
+from .data_collections import DataCollection
+from .data_collections_bands import Band, Unit
+
+DTYPE_TO_SAMPLE_TYPE: Dict[type, str] = {
+    bool: "UINT8",
+    np.uint8: "UINT8",
+    np.uint16: "UINT16",
+    np.float32: "FLOAT32",
+}
+
+EVALSCRIPT_TEMPLATE = """
+    //VERSION=3
+
+    function setup() {{
+        return {{
+            input: [{{
+                bands: [{input_names}],
+                units: [{input_units}]
+            }}],
+            output: [{output_spec}]
+        }}
+    }}
+
+    function updateOutputMetadata(scenes, inputMetadata, outputMetadata) {{
+        outputMetadata.userData = {{
+            "norm_factor":  inputMetadata.normalizationFactor
+        }}
+    }}
+
+    function evaluatePixel(sample) {{
+        return {{ {return_spec} }};
+    }}
+"""
+
+
+def parse_data_collection_bands(data_collection: DataCollection, bands: List[str]) -> List[Band]:
+    """Checks that all requested bands are available and returns the band information for further processing
+
+    :param data_collection: A collection of requested satellite data.
+    :param bands: A list of band or meta band names to use in the evalscript.
+    """
+
+    requested_bands = []
+    band_info_dict = {band_info.name: band_info for band_info in data_collection.bands + data_collection.metabands}
+
+    for band_name in bands:
+        if band_name not in band_info_dict:
+            raise ValueError(
+                f"Data collection {data_collection} does not have specifications for band {band_name}.\n"
+                f"Available bands are:\n{[band.name for band in data_collection.bands]}\nand meta-bands:\n"
+                f"{[band.name for band in data_collection.metabands]}."
+            )
+        requested_bands.append(band_info_dict[band_name])
+
+    return requested_bands
+
+
+def generate_evalscript(
+    data_collection: DataCollection,
+    bands: Optional[List[str]] = None,
+    meta_bands: Optional[List[str]] = None,
+    merged_output: Optional[str] = None,
+    use_dn: bool = True,
+) -> str:
+    """Generate an evalscript based on the provided specifications. This utility supports generating only evalscripts
+    with the mosaicking option set to `SIMPLE`.
+
+    :param data_collection: A collection of requested satellite data.
+    :param bands: A list of band names to use in the evalscript. Defaults to using all bands provided by the collection.
+    :param bands: A list of meta band names to use in the evalscript. By default no meta bands are added.
+    :param merged_output: If provided, bands will be concatenated into a single multi-band tiff with this name.
+    :param use_dn: Use DN units if possible. Default is True. If DN units are not available, the default units for each
+        specific band are used. DN units will be used regardless of the flag if they are the only possible choice.
+    """
+
+    band_names = bands if bands is not None else [band.name for band in data_collection.bands]
+    meta_band_names = meta_bands if meta_bands is not None else []
+
+    input_names, input_units, output_spec, return_spec = [], [], [], []
+    requested_bands = parse_data_collection_bands(data_collection, band_names + meta_band_names)
+    for band in requested_bands:
+        unit_choice = band.units.index(Unit.DN) if (use_dn and Unit.DN in band.units) else 0
+        sample_type = DTYPE_TO_SAMPLE_TYPE[band.output_types[unit_choice]]
+
+        input_names.append(f'"{band.name}"')
+        input_units.append(f'"{band.units[unit_choice].value}"')
+
+        # skip bands, since they will be provided as a single object
+        if merged_output is not None and band.name in band_names:
+            continue
+
+        # keep for meta_bands
+        output_spec.append(f'{{ id: "{band.name}", bands: 1, sampleType: "{sample_type}" }}')
+        return_spec.append(f"{band.name}: [sample.{band.name}]")
+
+    if merged_output is not None:
+        band = data_collection.bands[0]
+        unit_choice = band.units.index(Unit.DN) if (use_dn and Unit.DN in band.units) else 0
+        sample_type = DTYPE_TO_SAMPLE_TYPE[band.output_types[unit_choice]]
+        output_spec.append(f'{{ id: "{merged_output}", bands: {len(band_names)}, sampleType: "{sample_type}" }}')
+        return_spec.append(f"{merged_output}: [{', '.join(f'sample.{band}' for band in band_names)}]")
+
+    evalscript = EVALSCRIPT_TEMPLATE.format(
+        input_names=", ".join(input_names),
+        input_units=", ".join(input_units),
+        output_spec=", ".join(output_spec),
+        return_spec=", ".join(return_spec),
+    )
+
+    return evalscript
--- a/tests/test_evalscript.py
+++ b/tests/test_evalscript.py
@@ -1,0 +1,105 @@
+import json
+import re
+from typing import List, Optional, Tuple
+
+import pytest
+
+from sentinelhub import DataCollection
+from sentinelhub.data_collections_bands import Unit
+from sentinelhub.evalscript import generate_evalscript
+
+
+def parse_evalscript_info(evalscript: str) -> Tuple[List[str], List[str], List[Tuple[str, str, str]]]:
+    input_names_re = re.search(r"bands: (\[.*\])", evalscript)
+    input_units_re = re.search(r"units: (\[.*\])", evalscript)
+    output_spec = re.findall(r'\{ id: "([^\s.]+)", bands: ([^\s.]+), sampleType: "([^\s.]+)" \}', evalscript)
+
+    assert input_names_re is not None, "regex failed for extraction of input names"
+    assert input_units_re is not None, "regex failed for extraction of input units"
+
+    input_names = json.loads(input_names_re.group(1))
+    input_units = json.loads(input_units_re.group(1))
+
+    return input_names, input_units, output_spec
+
+
+@pytest.mark.parametrize(
+    "data_collection, bands",
+    [
+        (DataCollection.SENTINEL2_L1C, None),
+        (DataCollection.SENTINEL2_L1C, ["B04", "B03", "B02"]),
+        (DataCollection.SENTINEL1_IW, ["VV", "VH"]),
+    ],
+)
+def test_input_bands(data_collection: DataCollection, bands: Optional[List[str]]) -> None:
+    evalscript = generate_evalscript(data_collection=data_collection, bands=bands)
+    input_names, *_ = parse_evalscript_info(evalscript)
+    assert len(input_names) == len(bands) if bands is not None else len(data_collection.bands)
+
+
+@pytest.mark.parametrize(
+    "meta_bands",
+    [None, ["CLM"], ["CLP", "dataMask"]],
+)
+def test_input_meta_bands(meta_bands: Optional[List[str]]) -> None:
+    data_collection = DataCollection.SENTINEL2_L1C
+    evalscript = generate_evalscript(data_collection=data_collection, meta_bands=meta_bands)
+    input_names, *_ = parse_evalscript_info(evalscript)
+
+    num_meta_bands = 0 if meta_bands is None else len(meta_bands)
+    assert len(input_names) == len(data_collection.bands) + num_meta_bands
+
+
+@pytest.mark.parametrize("use_dn", [True, False])
+@pytest.mark.parametrize(
+    "data_collection, bands, meta_bands",
+    [
+        (DataCollection.SENTINEL2_L1C, ["B04", "B03", "B02"], ["CLP", "sunZenithAngles"]),
+        (DataCollection.SENTINEL3_OLCI, ["B04", "B03"], ["HUMIDITY"]),
+    ],
+)
+def test_input_units(data_collection: DataCollection, bands: List[str], meta_bands: List[str], use_dn: bool) -> None:
+    evalscript = generate_evalscript(data_collection=data_collection, bands=bands, meta_bands=meta_bands, use_dn=use_dn)
+    input_names, input_units, _ = parse_evalscript_info(evalscript)
+
+    band_info_dict = {band_info.name: band_info for band_info in data_collection.bands + data_collection.metabands}
+    for name, unit in zip(input_names, input_units):
+        band = band_info_dict[name]
+        assert Unit(unit) in band.units
+
+        if use_dn and Unit.DN in band.units:
+            assert Unit(unit) == Unit.DN
+
+
+@pytest.mark.parametrize(
+    "data_collection, bands, meta_bands",
+    [
+        (DataCollection.LANDSAT_TM_L2, None, ["BQA"]),
+        (DataCollection.SENTINEL2_L2A, ["B04", "B03", "B02"], ["CLP", "SCL"]),
+    ],
+)
+def test_merged_output(data_collection: DataCollection, bands: Optional[List[str]], meta_bands: List[str]) -> None:
+    merged_output = "merged_bands"
+    evalscript = generate_evalscript(
+        data_collection=data_collection, bands=bands, meta_bands=meta_bands, merged_output=merged_output
+    )
+    *_, output_spec = parse_evalscript_info(evalscript)
+
+    num_bands = len(data_collection.bands) if bands is None else len(bands)
+    depth = [depth for name, depth, _ in output_spec if name == merged_output][0]
+
+    assert len(output_spec) == 1 + len(meta_bands)
+    assert int(depth) == num_bands
+
+
+@pytest.mark.parametrize("merged_output", [None, "merged_bands"])
+@pytest.mark.parametrize("use_dn", [True, False])
+def test_sample_type(merged_output: Optional[str], use_dn: bool) -> None:
+    data_collection = DataCollection.SENTINEL2_L1C
+    evalscript = generate_evalscript(data_collection=data_collection, merged_output=merged_output, use_dn=use_dn)
+
+    *_, output_spec = parse_evalscript_info(evalscript)
+    assert len(output_spec) == 1 if merged_output else len(data_collection.bands)
+
+    sample_types_set = {sample_type for *_, sample_type in output_spec}
+    assert sample_types_set == {"UINT16" if use_dn else "FLOAT32"}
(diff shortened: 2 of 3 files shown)


mlubej:
Tested in place for `eo-learn` as well as `s2cloudless` and it works for both. Will wait for this MR before implementing changes in other places, to avoid going blindly in the wrong direction.

sentinelhub/evalscript.py
```diff
@@ -0,0 +1,117 @@
...
+            }}],
+            output: [{output_spec}]
+        }}
+    }}
+
+    function updateOutputMetadata(scenes, inputMetadata, outputMetadata) {{
+        outputMetadata.userData = {{
+            "norm_factor":  inputMetadata.normalizationFactor
+        }}
+    }}
```
zigaLuksic, line 35: will this always be available? even when not using `DN`?

sentinelhub/evalscript.py
```diff
@@ -0,0 +1,117 @@
...
+
+    return requested_bands
+
+
+def generate_evalscript(
+    data_collection: DataCollection,
+    bands: Optional[List[str]] = None,
+    meta_bands: Optional[List[str]] = None,
+    merged_output: Optional[str] = None,
+    use_dn: bool = True,
```
zigaLuksic, line 70: This only allows us to switch to `DN` but not other units right?
To be fair ATM it is very rare to even have two options for units, but this way the user can't get bands + `sunAzimuthAngles` because those only have degrees.

I know it's a weird edge case, but the way this is now it's not a direct replacement for `eo-learn`. Should we do a direct replacement? Or do we make something else and wait for new tasks to replace SHIT

sentinelhub/evalscript.py
```diff
@@ -0,0 +1,117 @@
...
+) -> str:
+    """Generate an evalscript based on the provided specifications. This utility supports generating only evalscripts
+    with the mosaicking option set to `SIMPLE`.
```
zigaLuksic, line 73: ah, this would also not work with the current SHIT API

sentinelhub/evalscript.py
zigaLuksic: ah, i see from the docs that it is more of a 'force_dn_when_possible`
I do like the idea going forward, just sad that SHIT is as it is :(

sentinelhub/evalscript.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+
+    band_names = bands if bands is not None else [band.name for band in data_collection.bands]
+    meta_band_names = meta_bands if meta_bands is not None else []
+
+    input_names, input_units, output_spec, return_spec = [], [], [], []
+    requested_bands = parse_data_collection_bands(data_collection, band_names + meta_band_names)
+    for band in requested_bands:
+        unit_choice = band.units.index(Unit.DN) if (use_dn and Unit.DN in band.units) else 0
+        sample_type = DTYPE_TO_SAMPLE_TYPE[band.output_types[unit_choice]]
```
zigaLuksic, line 90: this is the kind of bullshit that we could avoid by developing these evalscripts together with sh-py

tests/test_evalscript.py
```diff
@@ -0,0 +1,105 @@
...
+    input_units_re = re.search(r"units: (\[.*\])", evalscript)
+    output_spec = re.findall(r'\{ id: "([^\s.]+)", bands: ([^\s.]+), sampleType: "([^\s.]+)" \}', evalscript)
+
+    assert input_names_re is not None, "regex failed for extraction of input names"
+    assert input_units_re is not None, "regex failed for extraction of input units"
+
+    input_names = json.loads(input_names_re.group(1))
+    input_units = json.loads(input_units_re.group(1))
+
+    return input_names, input_units, output_spec
```
zigaLuksic, line 23: damn, you'd almost need to test that this works as intended :see_no_evil: 

could you replace some use of this with `assert expected_str in evalscript`?

tests/test_evalscript.py
```diff
@@ -0,0 +1,105 @@
...
+    evalscript = generate_evalscript(data_collection=data_collection, bands=bands, meta_bands=meta_bands, use_dn=use_dn)
+    input_names, input_units, _ = parse_evalscript_info(evalscript)
+
+    band_info_dict = {band_info.name: band_info for band_info in data_collection.bands + data_collection.metabands}
+    for name, unit in zip(input_names, input_units):
+        band = band_info_dict[name]
+        assert Unit(unit) in band.units
+
+        if use_dn and Unit.DN in band.units:
+            assert Unit(unit) == Unit.DN
```
zigaLuksic, line 71: seems quite programatic :thinking:

tests/test_evalscript.py
```diff
@@ -0,0 +1,105 @@
...
+    evalscript = generate_evalscript(
+        data_collection=data_collection, bands=bands, meta_bands=meta_bands, merged_output=merged_output
+    )
+    *_, output_spec = parse_evalscript_info(evalscript)
+
+    num_bands = len(data_collection.bands) if bands is None else len(bands)
+    depth = [depth for name, depth, _ in output_spec if name == merged_output][0]
+
+    assert len(output_spec) == 1 + len(meta_bands)
+    assert int(depth) == num_bands
```
zigaLuksic, line 92: gonna be honest with you, these tests have so much code that i am more sure you got the evalscript right than that you got the tests right :see_no_evil:

tests/test_evalscript.py
zigaLuksic, line 1: I'm missing a test where an evalscript like this would actually be used with the service :D
Doesn't need to be multiple, and just use a small area. We just want to see that the service does not reject them.
(dont forget to mark it with `sh_intergration`)

sentinelhub/evalscript.py
mlubej: Yes, the logic behind it is to always have it available, and the user uses this based on their specific needs, as in they should know what they're doing.

If the user doesn't specify the `userdata` in the response, then this doesn't get downloaded anyway.

sentinelhub/evalscript.py
mlubej: > ah, i see from the docs that it is more of a 'force_dn_when_possible`

Yes, correct, the flag means to `use DN if possible, otherwise the first default choice`. Do you think the name of the parameter is too misleading? Perhaps it could be `dn_priority` or `prioritize_dn` or something like that

> This only allows us to switch to DN but not other units right?

That seemed like a too specific case to me, so in this situation it would be up to the user to provide a custom defined evalscript.

> I know it's a weird edge case, but the way this is now it's not a direct replacement for eo-learn.

> , just sad that SHIT is as it is :(

Could you elaborate what you mean exactly? I see no issue with the above being the direct replacement for `eo-learn`. Or perhaps that was still the old comment.

sentinelhub/evalscript.py
mlubej: in SHIT this made sense only in cases where the provided evalscript was a custom one and not the autogenerated one, I believe. Not even sure why the custom evalscript is an option in the first place, isn't SHET supposed to take care of that functionality already?

sentinelhub/evalscript.py
mlubej: yep, felt like BS, but I guess it is the way it is.

tests/test_evalscript.py
mlubej: just to be clear, this is not a test, just a utility to extract test which is then later tested throughout the file.

In this case I would prefer to add some generic simple tests which don't rely on this function and check for simple substring presence

tests/test_evalscript.py
mlubej: not sure what exactly you mean, would you prefer `assert unit == Unit.DN.value`? perhaps that would be more explicit, to reduce extra wraps on the tested object

tests/test_evalscript.py
mlubej: yeah.. you should've seen what it looked like before, the tests were adapted from the s2cloudless evalscript generation test.

I'll take a second look and see what can be done to make them more intuitive, or maybe split them up additionally to use less if statements

sentinelhub/evalscript.py
zigaLuksic: it's not a direct replacement, it would feature some API changes
but i reckon that would be small enough to be worth it

sentinelhub/evalscript.py
zigaLuksic: but does the `normalizationFactor` exist if one downloads with REFLECTANCE?

sentinelhub/evalscript.py
zigaLuksic: no clue :upside_down_face: its why i'm so keen on replacing it with something better

what precisely would change for other mosaicking options?

tests/test_evalscript.py
zigaLuksic: it almost feels like you're reproducing the implementation here :thinking: 
can you perhaps just add 'expected_bands' to each test case? so that the expected output is part of the paramtrization, and you then just do `extracted_bands == expected_bands` or something similar, without all the conditionals and loops.

sentinelhub/evalscript.py
mlubej: Yes, because the info is taken from the metadata of the collection. I tried it for a couple of different collections as well, it's always set to some value and based on the docs always represents the factor in the formula for getting reflectances from DN.

sentinelhub/evalscript.py
mlubej: Ah, wait, I effed up. There is `mosaicking` and `mosaicking_order`:
- the first one controls the flow of data in the evalscript (either it's a single mosaic for a given time interval, or the samples are split per orbit/tile)
  - in generated evalscripts this is always set to SIMPLE, but in custom evalscripts this can be set to anything else, but not sure why this should be supported in SHIT, I would remove custom evalscripts from SHIT, since we have SHET for that.
- the `mosaicking_order` parameter only controls the order in which the data is tiled, (most recent, least recent, least cc..). this is not controlled via the evalscript, but provided in the request payload separately, so from this point of view nothing changes.

tests/test_evalscript.py
mlubej: I guess we could go about it in a different way, this just allowed me to work with objects instead of strings, but I guess if you have string comparison, everything else is more explicit and intuitive as you construct the string to compare.

I'll try it out

mlubej: managed to remove the utility

tests/test_evalscript.py
mlubej: tried to make them a bit more simple and to make direct comparisons and string inclusion checks. I guess it's more clear from the look of it what the test is trying to achieve, but there are still loops inside because the input/output approach to parametrization didn't work out as nice because of the longer arrays of band names

zigaLuksic: some looping is ok, as long as its not hard to understand the test

tests/test_evalscript.py
```diff
@@ -1,105 +1,100 @@
...
+    expected_uint_count = len(data_collection.bands) if use_dn else 0
+    assert evalscript.count('"UINT16"') == expected_uint_count
+
+    expected_float_count = len(data_collection.bands) if not use_dn else 0
+    assert evalscript.count('"FLOAT32"') == expected_float_count
+
+
+@pytest.mark.parametrize("use_dn", [True, False], ids=lambda x: f"use_dn: {x}")
+def test_sample_type_merged(use_dn: bool) -> None:
+    merged_output = "merged_bands"
```
zigaLuksic, line 92: what does this do?

mlubej: Generally I had this pattern for cases where I would be using the variable in multiple places, but I see here I just copied the usage. Will just move it inline.

mlubej: Yes. better name suggestions are welcome.

This functionality is sometimes needed when you don't want to download separate bands (in s2cloudless). It could even be useful in eo-grow, where we could remove the extra functionality to explode bands perhaps.

zigaLuksic: so this is both a flag and a name of the output file? gonna need to mull this over :thinking:

tests/test_evalscript.py
```diff
@@ -0,0 +1,136 @@
...
+    evalscript = generate_evalscript(
+        data_collection=data_collection,
+        bands=bands,
+        meta_bands=meta_bands,
+        merged_output=merged_output,
+        use_dn=use_dn,
+    )
+    if merged_output is not None:
+        responses = [SentinelHubRequest.output_response(merged_output, MimeType.TIFF)]
+        responses.extend([SentinelHubRequest.output_response(band, MimeType.TIFF) for band in meta_bands])
```
zigaLuksic, line 117: i see, this merges together bands, but not meta bands as well?

tests/test_evalscript.py
```diff
@@ -0,0 +1,136 @@
...
+    assert evalscript.count('"UINT16"') == expected_uint_count
+
+    expected_float_count = 1 if not use_dn else 0
+    assert evalscript.count('"FLOAT32"') == expected_float_count
+
+
+@pytest.mark.parametrize("data_collection", [DataCollection.LANDSAT_TM_L2, DataCollection.SENTINEL2_L2A])
+@pytest.mark.parametrize("merged_output", [None, "bands"])
+@pytest.mark.parametrize("use_dn", [True, False])
+def test_valid_evalscript(data_collection: DataCollection, merged_output: Optional[str], use_dn: bool) -> None:
```
zigaLuksic, line 105: instead of tagging all of the tests in the module, i think you can just slap it on here with `@pytest.mark.sh_integration` since other tests don't need SH

tests/test_evalscript.py
mlubej: indeed. meta bands are treated differently already in the data collection, so I thought it wouldn't make sense to treat them the same as bands here. I imagine it's more common to want to merge together bands separately. 

Additionally, this could then be simplified in eo-learn, where we request metabands separately, but ultimately manually pack the bands together into a single image. This code would now become redundant, because the band packing is done on the server side already.

tests/test_evalscript.py
mlubej: another option would be to use two parameters:
- `merge_bands_output: bool = False`
- `merged_bands_output_id: str = "bands"`

where the name is made parametrizable, but set to some default, so the user would more or less work only with the boolean flag. Again, accepting better names

tests/test_evalscript.py
zigaLuksic: So in `eo-learn` we download each band separately? What are the benefits of merging them together on the server side? does it cost less / takes up less space?

mlubej: A question for core, or perhaps @batic can help out, but what I do know is that it's common to pack the bands into a single output. One additional limitation is the JPEG/PNG only work with 1 or 3 bands, but here we use TIFF mimetype anyway, so it doesn't really make a difference from that point

tests/test_evalscript.py
mlubej: done, thanks!

tests/test_evalscript.py
batic: From the point of the service, no difference. Each band separately (e.g., `response["B01.tif"]`, `response["B02.tif"]` ...), or all the bands in one output (e.g. x-bands in `response["default.tif"]`) is the same. The only problem might come if one would pack tens (or hundreds) of bands into a single tiff. The PUs are computed based on the number of inputs. 

IIRC, we download bands separately in eo-learn to reduce the download size, as we could request `DN`s for most of the bands, and `bool` for masks, `uint`s for some other data etc, and later convert them to `float` and pack them into a single feature. It probably still doesn't make sense, as typically you'd join together bands of the same type (e.g. Sentinel-2 bands into one feature, masks into masks, angles etc into its own feature etc).

mlubej: > IIRC, we download bands separately in eo-learn to reduce the download size, as we could request DNs for most of the bands, and bool for masks, uints for some other data etc, and later convert them to float and pack them into a single feature. It probably still doesn't make sense, as typically you'd join together bands of the same type (e.g. Sentinel-2 bands into one feature, masks into masks, angles etc into its own feature etc).

With the changes proposed here, you are now able to download all bands as DN, already packed together into a single output, while still having other features (but only meta bands such as CLM, CLM, dataMask, ..) as separate response objects, so all good from this point of view as well.

thanks!

batic: Looking at [`S2-L1C`](https://docs.sentinel-hub.com/api/latest/data/sentinel-2-l1c/#units) band units, the angles (which are eo-learn `data` features), are not in DNs. Is this taken into account?

mlubej: Yes, the `use_dn` parameter is more of a `use_dn_if_possible_otherwise_the_first_choice`, which also covers the case of `dont_use_dn_unless_they_are_the_only_possible_choice`

accepting better name suggestions :D

zigaLuksic: `prefer_dn`?

mlubej: Yes, most likely something along those lines. I was also thinking about `prioritize_dn`, but it's longer

zigaLuksic: that one is good, i'd go with that. it's more formal :)

but i'm stil on the verge for merging bands. If its of no benefit to us, i'm not sure why we'd try and do it in the evalscript :thinking: it makes the evalscript function quite a bit more complex, so i'd want a trade-off

mlubej: > but i'm stil on the verge for merging bands. If its of no benefit to us, i'm not sure why we'd try and do it in the evalscript 🤔 it makes the evalscript function quite a bit more complex, so i'd want a trade-off

well, there's no benefit from the memory/speed/cost point of view, but there is a technical cost of manually packing the bands together. I think the code for supporting this in the evalscript is much nicer and less complex than the code needed to pack the actually data together manually:
- much of the code in [eo-learn](https://github.com/sentinel-hub/eo-learn/blob/develop/io/eolearn/io/sentinelhub_process.py#L593-L634) would become redundant due to this
- the [ExplodeBandsTask](https://github.com/sentinel-hub/eo-learn/blob/develop/core/eolearn/core/core_tasks.py#L517-L538) might become redundant as well at some point in the future, if we tweak the SHIT task in 4.0 so that it can also accept the option of providing bands separately

mlubej: > that one is good, i'd go with that. it's more formal :)

done

>  it makes the evalscript function quite a bit more complex, so i'd want a trade-off

I checked again and tried to make it a bit less complex/bloated and more intuitive in the last commit, please take a look when you manage

zigaLuksic: i actually like it more before, i don't like changing state in the for loops (unless it's appending or something) :see_no_evil: 

what if you gather the information for output in a separate list/datastructure. You need a `band_name: sample_type` map right? Then after this loop you branch on whether the output is merged or not, and either you do a `.join` or just an f-string. Ping me on mattermost if i'm unclear.

mlubej: agreed. did what you suggested and it's much more elegant now. I need to get used to seeing more clearly what is the obstacle and trying to extract the information beforehand, so that the flow later on is easier to follow.

thanks!

sentinelhub/evalscript.py
```diff
@@ -0,0 +1,118 @@
...
+
+    output_spec, return_spec = [], []
+    if merged_output is not None:
+        sample_type = sample_type_map[band_names[0]]
+        output_spec.append(f'{{id: "{merged_output}", bands: {len(band_names)}, sampleType: "{sample_type}"}}')
+        return_spec.append(f"{merged_output}: [{', '.join(f'sample.{band_name}' for band_name in band_names)}]")
+
+    for band_name, sample_type in sample_type_map.items():
+        if merged_output is not None and band_name in band_names:
+            continue
```
zigaLuksic, line 106: you can now switch the `for` and `if` statement and get rid of the `continue` right? In fact it can be an `else`.

tests/test_evalscript.py
```diff
@@ -0,0 +1,137 @@
...
+    assert expected_bands_str in evalscript
+
+
+@pytest.mark.parametrize(
+    "data_collection, bands",
+    [
+        (DataCollection.SENTINEL2_L2A, ["B04", "B03", "B02"]),
+        (DataCollection.SENTINEL1_IW, ["VV", "VH"]),
+    ],
+    ids=str,
```
zigaLuksic, line 21: wont this make really long names? perhaps the name can be the data collection?

tests/test_evalscript.py
zigaLuksic: perhaps a `merged_bands_output` is sufficient for this. It is more specific than `merged_output` so I suspect people will read the docs on what it does.

sentinelhub/evalscript.py
mlubej: This part of the code has to take care of the bands as well as meta-bands, it just excludes bands if the merged output is provided.

But otherwise you're right, if I change the condition, I could still get rid of the `continue`, if I do it like this:

```python
for band_name, sample_type in sample_type_map.items():
    if merged_bands_output is None or band_name in meta_band_names:
        output_spec.append(f'{{id: "{band_name}", bands: 1, sampleType: "{sample_type}"}}')
        return_spec.append(f"{band_name}: [sample.{band_name}]")
```

tests/test_evalscript.py
mlubej: It's not so long, and I prefer to know which test I'm running in vscode.

image

If you still think it's too ugly/long, I can revert

tests/test_evalscript.py
mlubej: agreed, done.

sentinelhub/evalscript.py
mlubej: added a comment, let me know if it's clearer now

zigaLuksic (approved):

PR #444, opened 2023-04-04
Move get_timestamps utility to sh-py

mlubej:
Changes:
- port the utility function and update it a bit
- add tests

Added to Catalog API because it basically relies upon it, so I see it as an extra utility built on top of the class.

### DIFF
--- a/sentinelhub/api/catalog.py
+++ b/sentinelhub/api/catalog.py
@@ -4,10 +4,13 @@ A client interface for `Sentinel Hub Catalog API <https://docs.sentinel-hub.com/
 import datetime as dt
 from typing import Any, Dict, Iterable, List, Literal, Optional, Union
 
+import numpy as np
+
 from ..base import FeatureIterator
+from ..config import SHConfig
 from ..data_collections import DataCollection, OrbitDirection
 from ..geometry import CRS, BBox, Geometry
-from ..time_utils import parse_time, parse_time_interval, serialize_time
+from ..time_utils import filter_times, parse_time, parse_time_interval, serialize_time
 from ..types import JsonDict, RawTimeIntervalType, RawTimeType
 from .base import SentinelHubService
 from .utils import remove_undefined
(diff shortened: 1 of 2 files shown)


sentinelhub/api/catalog.py
```diff
@@ -265,3 +268,47 @@ def get_ids(self) -> List[str]:
...
+    :return: A list of timestamps of available observations.
+    """
+    query_filter = None
+    time_difference = time_difference if time_difference is not None else dt.timedelta(seconds=-1)
+    fields = {"include": ["properties.datetime"], "exclude": []}
+
+    if maxcc is not None and data_collection.has_cloud_coverage:
+        if np.issubdtype(type(maxcc), np.integer):
+            raise ValueError('Maximum cloud coverage "maxcc" parameter should be a float on an interval [0, 1]')
+        query_filter = f"eo:cloud_cover < {int(maxcc * 100)}"
```
mlubej, line 302: Not quite sure about raising this error, are the types not enough in this case? is it better to notify the user explicitly to make sure they're using the parameter the right way?

zigaLuksic: the types don't cover the [0, 1] info. Someone might try to put in percentages.

mlubej: constrained the check back to the original.

sentinelhub/api/catalog.py
```diff
@@ -265,3 +266,47 @@ def get_ids(self) -> List[str]:
...
+
+    if maxcc is not None and data_collection.has_cloud_coverage:
+        if isinstance(maxcc, (int, float)) and (maxcc < 0 or maxcc > 1):
+            raise ValueError('Maximum cloud coverage "maxcc" parameter should be a float on an interval [0, 1]')
+        query_filter = f"eo:cloud_cover < {int(maxcc * 100)}"
+
+    if data_collection.service_url is not None:
+        config = config.copy() if config else SHConfig()
+        config.sh_base_url = data_collection.service_url
+
```
zigaLuksic, line 305: this is just so ugly :( but i guess we leave it for now

zigaLuksic (approved):

sentinelhub/api/catalog.py
mlubej: does anything change with the new config version? I guess I could put the url in the config init, but it doesn't change much.

I wonder what would be the proper way to go about such things.

zigaLuksic: signatures would need to be changed, 4.0 perhaps

PR #449, opened 2023-04-18
Add y-labels to statistical example

zigaLuksic:
I am unsure why the diff decided everything was changed...
It is actually just two lines (and two images), which specify the y-label, which Anja Vrečko noticed was [missing](https://sentinelhub-py.readthedocs.io/en/latest/examples/statistical_request.html) (bottom example).
The y-label is now `NDVI mean`.


mlubej:
> I am unsure why the diff decided everything was changed... It is actually just two lines (and two images)

Images are rendered as a strings in the diff, and the amount of change is probably over Github's threshold of showing details.

mlubej (approved):

zigaLuksic:
> > I am unsure why the diff decided everything was changed... It is actually just two lines (and two images)
> 
> Images are rendered as a strings in the diff, and the amount of change is probably over Github's threshold of showing details.

It actually worked fine until i reverted some changes in VSCode :upside_down_face: so weird...

PR #451, opened 2023-04-24
Generalize Geometry class to better leverage shapely for parsing values

zigaLuksic:
Had to inspect the shapely code to get the errors a bit more understandable (attribute errors are hard to decipher in such cases).

Tested that it works with `gpkg` now (as you suggested).

### DIFF
--- a/sentinelhub/geometry.py
+++ b/sentinelhub/geometry.py
@@ -11,6 +11,7 @@ import shapely.geometry
 import shapely.geometry.base
 import shapely.ops
 import shapely.wkt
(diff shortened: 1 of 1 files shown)


batic:
Yeah, this is (temporary) solution, imo as well. 

In principle, a `Geometry.parse` method that doesn't rely on `__init__` would be better. There we could perhaps make sure things work with `wkt`, `wkb` and anything with geo interface (i.e., `shapely.geometry.shape`).

Approved.

batic (approved):

PR #452, opened 2023-04-26
Official Python 3.11 support

zigaLuksic:
Extending enums is problematic in 3.11, probably because the official implementation of Enums got upgraded in some ways.
If we use the class from `aenum` then it seems to work. Tests in eo-grow and eo-learn pass with this as well.

### DIFF
--- a/.github/workflows/ci_action.yml
+++ b/.github/workflows/ci_action.yml
@@ -67,6 +67,7 @@ jobs:
         python-version:
           - "3.9"
           - "3.10"
+          - "3.11"
         include:
           # A flag marks whether full or partial tests should be run
           # We don't run integration tests on pull requests from outside repos, because they don't have secrets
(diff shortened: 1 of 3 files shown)


mlubej (approved):
So, just to understand, the `enum` package got upgraded which caused issues, but the current fix is to use the same classes from `aenum`, where the upgrades aren't present, and therefore it still works?

zigaLuksic:
> So, just to understand, the `enum` package got upgraded which caused issues, but the current fix is to use the same classes from `aenum`, where the upgrades aren't present, and therefore it still works?

I suspect the upgrades to enum caused some issues. I have no clue, because it is a very weird interaction with how ray uses cloudpickle etc. When trying to build a minimal example it shouldn't ever work :D

I also have no clue what the diff between `enum.Enum` and `aenum.Enum` is, but both are from the same creator. And it seems like `aenum.Enum` and `aenum.extend_enum` are more in sync.

But yeah, we're abusing the concept of enums terribly here.

PR #453, opened 2023-05-03
Adjust auth url parameter of SHConfig to add more flexibility w.r.t. endpoints

zigaLuksic:


sentinelhub/config.py
```diff
@@ -25,7 +28,8 @@ class _SHConfig:
     sh_client_id: str = ""
     sh_client_secret: str = ""
     sh_base_url: str = "https://services.sentinel-hub.com"
-    sh_auth_base_url: str = "https://services.sentinel-hub.com"
+    sh_auth_base_url: Optional[str] = None
+    sh_auth_url: str = "https://services.sentinel-hub.com/oauth/token"
```
batic, line 32: Would perhaps `sh_token_url` be better name?

batic (approved):

PR #460, opened 2023-05-08
Ci cache tweaks

zigaLuksic:
Adjust false 3.8 cache caused by pre-commit, cache whole environment in other cases


zigaLuksic:
Got a lot of insight from this, closing to create a new minimal one

PR #469, opened 2023-05-24
Switch to ruff

zigaLuksic:
This might need some corrections, but i am having issues with the test suite locally (the OGC part takes more than 10 minutes and i'm getting weird dataclass errors)


mlubej (approved):
More or less the same things as in eo-learn, as discussed. Went through the commits anyway just in case but didn't find anything standing out.

PR #479, opened 2023-08-22
Simplify docs

zigaLuksic:
Removes all content that is not really relevant for us

had to hardcode the sphinx version because otherwise it would install a version under 2 (yeah) and autodocs wouldnt work


batic (approved):
🙇 👍

PR #486, opened 2023-11-02
Config no longer falsely loads profiles if file doesn't exist

zigaLuksic:
As noticed by you, if the file doesn't exist one could do the first initialization with `SHConfig("something")` and break the whole damn thing.

I added a test that now the first init fails for any non-default profile (since no such profile exists).


jonasViehweger:
Looks good, thanks for fixing

PR #487, opened 2023-11-02
Make sh-config CLI use environmental variable

zigaLuksic:
~the saving is also unnecessary because the load method creates the file if needed~

saving is needed because we have one do-all command. Gotta figure this out


mlubej (approved):
Confirming it works now. Really shows what a good refactor can do, a really elegant fix.

PR #493, opened 2023-11-13
Fix bug that occurs with dataclass_json 0.62

zigaLuksic:
The issue happens when someone calls the `to_dict` method on the collection.
Sadly this is used in our code a lot :upside_down_face: would you consider this worthy of a release?

### DIFF
--- a/sentinelhub/geometry.py
+++ b/sentinelhub/geometry.py
@@ -6,7 +6,7 @@ import contextlib
 import warnings
 from abc import ABCMeta, abstractmethod
 from math import ceil
-from typing import Callable, Dict, Iterator, Tuple, TypeVar, Union, cast
+from typing import Callable, Dict, Iterator, Tuple, TypeVar, Union
 
 import shapely.geometry
 import shapely.geometry.base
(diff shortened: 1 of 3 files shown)


zigaLuksic:
i guess a release makes sense also from the point that eo-grow tests get fixed. Otherwise we have to skip the byoc pipeline tests for a bit.

sentinelhub/geometry.py
```diff
@@ -159,9 +159,9 @@ def _tuple_from_list_or_tuple(
         :raises: TypeError
         """
         if len(bbox) == 4:
-            min_x, min_y, max_x, max_y = cast(Tuple[float, float, float, float], bbox)
+            min_x, min_y, max_x, max_y = bbox
         else:
-            (min_x, min_y), (max_x, max_y) = cast(Tuple[Tuple[float, float], Tuple[float, float]], bbox)
+            (min_x, min_y), (max_x, max_y) = bbox
         return float(min_x), float(min_y), float(max_x), float(max_y)
```
mlubej, line 165: Would it make sense to limit the various options how to define a bbox? it seems like a similar situation as in case of features in eolearn. But this is a separate issue, so this is just a comment.

mlubej (approved):
A minor release seems completely suitable to me.

sentinelhub/geometry.py
zigaLuksic: we already did quite a bit
the problem with that is that it is fairly code-breaking, so it's a sin of the past that we're slowly correcting

PR #494, opened 2023-11-13
Deprecate the deprecator

zigaLuksic:
The deprecator from typing_extensions has additional benefits of IDEs picking it up

image

I wanted to deprecate our own deprecators, but then users would get spammed with stuff that they have no control over (since it would happen on import. So I want to wait with that (or just remove them after a while)


mlubej (approved):
You plan to do a separate MR for preparing the release, right?

zigaLuksic:
> You plan to do a separate MR for preparing the release, right?

yes

PR #496, opened 2023-11-13
Further remove AWS docs

zigaLuksic:


examples/batch_statistical.ipynb
mlubej, line 1: I assume this contains the parts where aws credentials are provided but not used anywhere?

zigaLuksic: we probably changed this a while ago but didnt notice that it had to be changed in the example

mlubej (approved):

PR #505, opened 2023-12-07
Default retry time

zigaLuksic:
I added a parameterized default retry time, which we can then set in all services.

I also already set a class parameter in the `SentinelHubService` base class which the subclasses just override.

### DIFF
--- a/sentinelhub/download/sentinelhub_client.py
+++ b/sentinelhub/download/sentinelhub_client.py
@@ -34,7 +34,7 @@ class SentinelHubDownloadClient(DownloadClient):
     _CACHED_SESSIONS: ClassVar[dict[tuple[str, str], SentinelHubSession]] = {}
     _UNIVERSAL_CACHE_KEY = "universal-user", "default-url"
 
-    def __init__(self, *, session: SentinelHubSession | None = None, default_retry_time: float = 0, **kwargs: Any):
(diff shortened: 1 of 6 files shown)


sentinelhub/download/sentinelhub_client.py
```diff
@@ -84,10 +87,9 @@ def _execute_download(self, request: DownloadRequest) -> DownloadResponse:
                 )
                 response = self._do_download(request)
 
-                self._execute_thread_safe(self.rate_limit.update, response.headers)
-
                 if response.status_code == requests.status_codes.codes.TOO_MANY_REQUESTS:
                     warnings.warn("Download rate limit hit", category=SHRateLimitWarning)
+                    self._execute_thread_safe(self.rate_limit.update, response.headers, default=self.default_retry_time)
```
zigaLuksic, line 92: The thing i am not 100% sure about is this.
Currently `rate_limit.update` will try and get a retry number from headers, if it cannot find it, it will take 0. This is updated no matter what kind of response one gets.

Since we now add a default retry time, that should probably only go in effect when we get a 429 right? But do we ever get retry headers in a non-429 response? if yes, then i need to tweak the logic a bit more.

The only scenario (that i am not sure is realistic) is that a user makes a successful request, which has a retry time, which is not applied. They have an external for-loop and immediately call the client again, but the retry time is not applied. But tbh it does not seem like a valid scenario.

batic: cc @tomislijepcevic

PR #509, opened 2023-12-08
Update session caching according to core auth changes

zigaLuksic:
I was thinking of making this a major release since older versions might no longer cache sessions correctly.

### DIFF
--- a/tests/download/test_session.py
+++ b/tests/download/test_session.py
@@ -48,10 +48,7 @@ def test_session(session: SentinelHubSession) -> None:
 
 @pytest.mark.sh_integration()
 def test_token_info(session: SentinelHubSession) -> None:
-    info = session.info()
-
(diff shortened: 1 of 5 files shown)


tests/download/test_session.py
```diff
@@ -50,7 +50,7 @@ def test_session(session: SentinelHubSession) -> None:
 def test_token_info(session: SentinelHubSession) -> None:
     info = session.info()
 
-    for key in ["sub", "aud", "jti", "exp", "name", "email", "sid", "org", "did", "aid", "d"]:
+    for key in ["exp", "iat", "jti", "iss", "sub", "typ", "azp", "scope", "account"]:
```
batic, line 53: Should we only check for what we need/use, i.e. "azp"?

zigaLuksic: done

mlubej (approved):
Didn't follow the oauth changes discussions, so not sure how much I can contribute, but I don't see anything weird going on

PR #512, opened 2023-12-12
Feat/cdse reference

chorng:
* Update service url selection logic for the `get_available_timestamps` helper function, ogc, wfs, `TileSplitter`, processing api, and statistical api.
* Update tests

### DIFF
--- a/sentinelhub/api/base_request.py
+++ b/sentinelhub/api/base_request.py
@@ -8,7 +8,7 @@ from abc import ABCMeta, abstractmethod
 from typing import Any
 
 from ..base import DataRequest
-from ..constants import MimeType, MosaickingOrder, RequestType, ResamplingType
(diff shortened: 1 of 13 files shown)


chorng:
@zigaLuksic Seems that the pipeline failed because of the missing [cdse profile](https://github.com/sentinel-hub/sentinelhub-py/pull/512/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R37) which I used to run tests locally.

How could I add a profile to the pipeline for testing on CDSE?

Adrok24:
I havent review everything in detail yet. But maybe after making the tests pass by adding the instance_id for cdse, we should think how to add a flag to skip the cdse cases. Maybe a decorator to wrap test functions, which ignore shconfig(cdse) when a command like --skip-cdse is passed.

zigaLuksic:
closing this one for now

PR #514, opened 2023-12-18
Feat/cdse doc

chorng:
Update examples for Copernicus Data Space Ecosystem users:
- processing api
- catalog api
- ogc api
- large area utilities
- statistical api


examples/data/interpolation_evalscript_cdse.js
```diff
@@ -0,0 +1,177 @@
...
+    while (cDate < end) {
+        sampled_dates.push(cDate);
+        cDate = increase(cDate, period, period_unit);
+    }
+    return sampled_dates;
+}
+
+function is_valid(smp) {
+  // Check if the sample is valid (i.e. contains no clouds or snow)
+  let scl = smp.SCL;
```
Adrok24, line 116: Give me context, these cdse evalscripts are included because some bands are different?

examples/data_search_cdse.ipynb
Adrok24, line 1: Maybe including a line when you set config with the cdse base-url? or is this somewhere else?

examples/large_area_utilities_cdse.ipynb
Adrok24, line 1: Same comment as before

examples/ogc_request_cdse.ipynb
Adrok24, line 1: Here the same, but also when i click the link in the cell that explains the shconfig https://sentinelhub-py.readthedocs.io/en/latest/configure.html and there is no mention on CDSE endpoint. correct me if i am wrong

Adrok24 (commented):
In general if the examples work with the cdse endpoint it seems ok! but with the comments I made adding info about the CDSE, or showing where you set the endpoint. Maybe this is done somewhere else and i m not seeing it.

examples/data/interpolation_evalscript_cdse.js
chorng: Yes, the example was using CLM which does not exist in CDSE

examples/data_search_cdse.ipynb
chorng: It's going to be added to the configuration section

examples/ogc_request_cdse.ipynb
chorng: Configuration will be added to the configuration page to which you point

docs/source/configure.rst
```diff
@@ -96,6 +96,32 @@ Batch, BYOC, and other APIs). There is "OAuth clients" frame where we can create
...
+
+    config = SHConfig()
+    config.sh_client_id = 'oauth-client-id'
+    config.sh_client_secret = 'oauth-client-secret'
+    config.sh_base_url = 'https://sh.dataspace.copernicus.eu'
+    config.sh_token_url = 'https://identity.dataspace.copernicus.eu/auth/realms/CDSE/protocol/openid-connect/token'
+
+
+.. admonition:: Supported Data Collections
+
```
zigaLuksic, line 120: here I'd also mention that when using `DataCollections` they also need to be configured for CDSE and then a link to the appropriate example notebook.

docs/source/index.rst
```diff
@@ -12,4 +12,5 @@ Documentation of `sentinelhub` Python package
     logging
     examples
     tutorials
+    cdse
```
zigaLuksic, line 15: should we group these three together or is it fine the way it is now?

examples/data_search_cdse.ipynb
zigaLuksic, line 1: there are a few examples here which seem nearly identical to non-cdse ones. What are the relevant changes in these notebooks?

docs/source/configure.rst
chorng: It will be a bit tricky to mention it here. In short, APIs with a dedicated class, e.g., catalog, batch, batch stat, etc., do not require a specific `DataCollection.define_from()` configuration. Other requests built on `SentinelHubServices`, `SentinelHubFeatureIterator` and `TileSplitter` require the special configuration.

To  avoid confusion, I'd keep CDSE as a separate session and direct CDSE to follow examples there.

docs/source/index.rst
zigaLuksic: I was thinking of adding a .rst files named `Examples` which then includes these three subcategories (so that the sidebar is nicely grouped) unless you want the CDSE examples to be extra visible

docs/source/configure.rst
zigaLuksic: I was thinking something along the lines of 
```
 With the Copernicus Data Space Ecosystem Configuration, please find the available data collections on
    `Copernicus Data Space Ecosystem indexed in Sentinel Hub`_. See [the example notebook](link-to-notebook) on how to configure them for direct use in `sentinelhub-py`.
```

or do you think its too specific to process API and the likes?

chorng: How about this?
```
See [the Copernicus Data Space Ecosystem tutorials](link-to-cdse-index) on how to configure them for direct use in `sentinelhub-py`.
```

zigaLuksic: sounds good

examples/data_search_cdse.ipynb
zigaLuksic: This examples add about 6k lines to the repo, and only a fraction of these is really relevant.

The reason i see for adding these notebooks (as they are) is if CDSE is a priority and we want users to have examples that they can run end-2-end instead of running a regular example with some adjustments that we tell them how to do.

So at the end of the day, it depends on how important CDSE is for us, which I guess is 'very'. In that case I think we can proceed.

docs/source/configure.rst
chorng: Added

examples/data_search_cdse.ipynb
chorng: Yes I think we can proceed. The naming can be changed if we find users are still confused

zigaLuksic:
@chorng any other plans for this?

chorng:
@zigaLuksic Nope. I think we can merge this pr if you also agree with it.

PR #527, opened 2024-04-24
Add configuration to limit rate limit retries

Regan-Koopmans:
At the moment the default behaviour is to retry requests indefinitely when rate limits are reached on the server side. For some use-cases it would be helpful to limit this behaviour, and allow the calling code to handle this with some backoff mechanism. In particular, if multiple processes are submitting requests it does not make sense for them to compete over the finite rate limit.

To achieve this I have added a new config called `max_rate_limit_retries`, which is `None` by default. If this configuration has a numeric value we will raise an `OutOfRequestsException` once the number of download attempts has exceeded the configuration value. I am open to changing the naming if necessary.

### DIFF
--- a/sentinelhub/config.py
+++ b/sentinelhub/config.py
@@ -46,6 +46,7 @@ class _SHConfig:
     download_sleep_time: float = 5.0
     download_timeout_seconds: float = 120.0
     number_of_download_processes: int = 1
+    max_rate_limit_retries: int | None = None
 
     def __post_init__(self) -> None:
         if self.sh_auth_base_url is not None:
@@ -94,6 +95,7 @@ class SHConfig(_SHConfig):
           attempt this number exponentially increases with factor `3`.
         - `download_timeout_seconds`: Maximum number of seconds before download attempt is canceled.
         - `number_of_download_processes`: Number of download processes, used to calculate rate-limit sleep time.
+        - `max_rate_limit_retries`: Maximum number of retries caused by rate limits until an exception is raised.
 
     The location of `config.toml` for manual modification can be found with `SHConfig.get_config_location()`.
     """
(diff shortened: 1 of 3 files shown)


zigaLuksic:
I switched the base to `develop`
I also see that there are some issues with the tests ATM, i'll investigate

zigaLuksic:
@Regan-Koopmans  after the switch to `develop` it is a bit harder to see what the actual changes are. Could you perhaps rebase your changes to `develop`?

sentinelhub/config.py
```diff
@@ -46,6 +46,7 @@ class _SHConfig:
     download_sleep_time: float = 5.0
     download_timeout_seconds: float = 120.0
     number_of_download_processes: int = 1
+    max_rate_limit_retries: int | None = None
```
zigaLuksic, line 49: since this parameter covers any retriable error, it might as well be renamed to `max_retries`

zigaLuksic (commented):
Looks great. Ready for merging after renaming.

Do you have any preferences on when this should be released?

sentinelhub/config.py
Regan-Koopmans: That is true, I have pushed a change for this.

Regan-Koopmans:
> Do you have any preferences on when this should be released?

We have a soft launch next week. In an ideal world we could incorporate this change before then, but we would not need to block on this, as it is considered a "nice to have". We can also already develop against this change being released, since we know what exception to expect.

zigaLuksic:
> > Do you have any preferences on when this should be released?
> 
> We have a soft launch next week. In an ideal world we could incorporate this change before then, but we would not need to block on this, as it is considered a "nice to have". We can also already develop against this change being released, since we know what exception to expect.

We'll try to find time to release this week, but no promises :)

Regan-Koopmans:
Also thanks for the quick turnaround time with reviews!

PR #536, opened 2024-07-31
Wait for batch process API job to update status from PROCESSING to something else

mlubej:
A simple addition with more descriptions.

Another option would be to run the `while` loop as long as the status is `PROCESSING`, but perhaps the user would be confused why the job doesn't move forward as soon as the progress bars hit 100%. An additional reason is that the suggested implementation is perhaps more informative

### DIFF
--- a/sentinelhub/api/batch/utils.py
+++ b/sentinelhub/api/batch/utils.py
@@ -93,6 +93,13 @@ def monitor_batch_job(
     failed_tiles_num = finished_count - success_count
     if failed_tiles_num:
         LOGGER.info("Batch job failed for %d tiles", failed_tiles_num)
+
+    LOGGER.info("Waiting on batch job status update.")
+    while batch_request.status is BatchRequestStatus.PROCESSING:
+        time.sleep(sleep_time)
+        batch_request = batch_client.get_request(batch_request)
+
+    LOGGER.info("Batch job finished with status %s", batch_request.status.value)
     return tiles_per_status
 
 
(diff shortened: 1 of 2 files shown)


zigaLuksic (approved):

mlubej:
@zigaLuksic in the `monitor_batch_job` we have the `monitor_batch_analysis` function which waits until the batch analysis is finished and then moves on. Then we put the responsibility of starting the job on the user, but if the user doesn't start the batch job, the loop will run forever. 

Wouldn't it make more sense if the `monitor_batch_job` finishes early if the last user action was not `START`, and if it was, it waits for the `PROCESSING` status?

zigaLuksic:
> @zigaLuksic in the `monitor_batch_job` we have the `monitor_batch_analysis` function which waits until the batch analysis is finished and then moves on. Then we put the responsibility of starting the job on the user, but if the user doesn't start the batch job, the loop will run forever.
> 
> Wouldn't it make more sense if the `monitor_batch_job` finishes early if the last user action was not `START`, and if it was, it waits for the `PROCESSING` status?
> 
> Asking because some test cases don't make much sense

Because `monitor_batch_job` is meant to monitor a started batch job. I'm not sure we want to think about all the possible cases of it being used wrong, because there are too many. Or do you think it's crystal clear?

mlubej:
Fixed the tests by additionally mocking the `batch_client.get_request` calls and updating the number of expected calls/loops

> I'm not sure we want to think about all the possible cases of it being used wrong, because there are too many. Or do you think it's crystal clear?

It could very well be that I have a focus on a specific and this wouldn't hold in general, so I'm fine with leaving it like it is, I just wanted to point out that the tests simulate cases which don't seem to be realistic. I agree that it's hard to draw the line between the functionality and what is expected from the user. I guess that so far this has not been a problem, so let's leave it as it is.

PR #549, opened 2024-11-07
Add wait wrapper around Batch V1/V2 user action calls until the status changes

mlubej:
- Added a simple wrapper which first saves the old status before performing the action, then it repeatedly checks for a status update until the wait is over. if the status doesn't change, it logs a warning
- added also for Batch V2
- added mocked tests for both

### DIFF
--- a/sentinelhub/api/batch/process.py
+++ b/sentinelhub/api/batch/process.py
@@ -11,7 +11,7 @@ import time
 from dataclasses import dataclass, field
 from enum import Enum
 from functools import wraps
-from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, TypeVar, Union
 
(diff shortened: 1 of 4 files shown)


sentinelhub/api/batch/process.py
```diff
@@ -29,6 +31,23 @@
 BatchCollectionType = Union[str, dict, "BatchCollection"]
 
 
+def batch_user_action_wait_for_status_change(func: Callable) -> Callable:
+    @wraps(func)
+    def retrying_func(self: SentinelHubBatch, batch_request: BatchRequest) -> Json:
```
mlubej, line 36: @zigaLuksic not sure about the `self` type here. It works for batch V1, but batch V2 in principle has a different class, which I don't import here, but `mypy` seems to be happy

mlubej:
Closing in favor of https://github.com/sentinel-hub/eo-grow/pull/360

PR #564, opened 2025-09-09
Fix tests in develop

mlubej:


examples/fis_request.ipynb
zigaLuksic, line 1: actually FIS is being deprecated, so it's better to remove this notebook completely IMO

mlubej: done

PR #565, opened 2025-09-10
Remove batch processing v1

zigaLuksic:
I was looking for a way to add a warning or error if anyone tries to import version 1, but since we reexport everything in upper modules i don't think there is a nice way to do that

### DIFF
--- a/sentinelhub/api/batch/process_v2.py
+++ b/sentinelhub/api/batch/process_v2.py
@@ -4,7 +4,7 @@ Deprecated module for Batch Processing v2 API. The contents have been moved to `
 
 import warnings
 
-from sentinelhub.api.batch.process import *  # noqa: F403
+from sentinelhub.api.batch.process import *  # noqa: F403 # pylint: disable=unused-wildcard-import, wildcard-import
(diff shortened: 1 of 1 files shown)


sentinelhub/api/batch/base.py
```diff
@@ -36,9 +34,8 @@ class BatchUserAction(Enum):
 
     START = "START"
     ANALYSE = "ANALYSE"
-    NONE = "NONE"
-    CANCEL = "CANCEL"
     STOP = "STOP"
+    NONE = "NONE"
```
mlubej, line 38: I only see this one used in the batch statistical notebook, but I don't see it documented in the official docs. Should it be removed?

zigaLuksic: I also didn't see it in the official docs, but when i removed it the tests failed because the enum couldn't parse this value so I added it back in

examples/batch_processing_v2.ipynb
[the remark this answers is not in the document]
zigaLuksic: renamed the code file and added a redirection warning

image

PR #582, opened 2026-09-11
Fix session sharing on Windows

### LINKED ISSUE #573: [BUG] SessionSharing doesn't work on Windows - shared memory destroyed after handle closed
**Describe the bug**

SessionSharingThread creates shared memory but immediately closes the handle in _share_token(). On Windows, shared memory is destroyed when all handles are closed, so worker processes cannot access it.

**To Reproduce**
## Environment

- **OS:** Windows 10/11
- **Python:** 3.12
- **sentinelhub-py:** 3.11.x

## Steps to Reproduce
```python
from concurrent.futures import ProcessPoolExecutor
from sentinelhub import SentinelHubSession, SHConfig
from sentinelhub.download import SessionSharing, collect_shared_session

def worker():
    session = collect_shared_session()
    return session.token["access_token"][:20]

config = SHConfig()
# config.sh_client_id = "..."
# config.sh_client_secret = "..."

session = SentinelHubSession(config)

with SessionSharing(session), ProcessPoolExecutor(max_workers=1) as executor:
    future = executor.submit(worker)
    print(future.result())
```

## Expected Behavior

Worker processes should be able to access the shared session token.

## Actual Behavior
```
FileNotFoundError: [WinError 2] The system cannot find the file specified: 'sh-session-token'
```

## Root Cause

In `_share_token()`, the shared memory handle is closed after writing:
```python
def _share_token(self, token: JsonDict) -> None:
    encoded_token = json.dumps(token).encode()
    memory = self._get_shared_memory(encoded_token)

    try:
        memory.buf[:] = encoded_token + _NULL_MEMORY_VALUE * (memory.size - len(encoded_token))
    finally:
        memory.close()  # <-- This destroys the memory on Windows

dajiaohuang:
`SessionSharingThread` closed the shared memory handle right after writing the token into it. On Windows a named shared memory block is destroyed as soon as the last handle to it is closed, so no other process could read the token and `collect_shared_session` raised `FileNotFoundError`.

The memory object is now kept open until the thread is joined, where it is unlinked and closed. A pre-existing block that cannot be removed because another thread or process still holds it, which is always the case on Windows, is reused instead of being allocated anew.

Added a test that keeps the thread sharing refreshed tokens. The session sharing tests fail on Windows before this change and pass after it; validation was run on Windows with Python 3.13 because the failure is Windows-specific.

Fixes #573


zigaLuksic:
Hi @dajiaohuang, thank you for the contribution. Currently sh-py is in maintenance mode, so we cannot say when we will have the time to fully test and adopt your solution.

PR #583, opened 2026-09-11
docs: correct the profile environment variable name in SHConfig docstring

dajiaohuang:
## Summary

`SHConfig.__init__` documents the profile environment variable as `SH_USER_PROFILE`, but that variable does not exist. The class actually reads `SH_PROFILE` (`sentinelhub/config.py:21` defines `SH_PROFILE_ENV_VAR = "SH_PROFILE"` and `_get_profile` looks it up at `sentinelhub/config.py:147`).

The stale name has been in the docstring since the environment-variable support was added, and it is rendered in the published reference documentation (`docs/source/conf.py` runs `sphinx-apidoc` over the whole `sentinelhub` package, excluding only the `commands.py` modules, with `autoclass_content = "both"`), so users following it would set a variable that silently has no effect.

`docs/source/configure.rst` and `CHANGELOG.MD` already use the correct `SH_PROFILE` name, so this only brings the API docstring in line with them.

## Changes

- `sentinelhub/config.py`: `SH_USER_PROFILE` -> `SH_PROFILE` in the `SHConfig.__init__` docstring.

## Test plan

- [x] `pytest tests/test_config.py` — 14 passed. This includes the existing test that sets `SH_PROFILE_ENV_VAR` and asserts that `SHConfig` picks up the profile from the environment, which is the behavior the corrected sentence describes.
- [x] `black --check`, `ruff check`, `pylint` on `sentinelhub/config.py` — clean.

### DIFF
--- a/.github/workflows/scheduled_caller.yml
+++ b/.github/workflows/scheduled_caller.yml
@@ -3,7 +3,7 @@ name: scheduled build caller
 on:
   schedule:
     # Schedule events are triggered by whoever last changed the cron schedule
-    - cron: "0 0 * * *"
+    - cron: "0 0 1 * *"
 
 jobs:
   call-workflow:
--- a/CHANGELOG.MD
+++ b/CHANGELOG.MD
@@ -1,3 +1,7 @@
+## [Version 3.11.4] - 2026-02-04
+
+- Remove API for deprecated FIS service
+
 ## [Version 3.11.3] - 2025-10-07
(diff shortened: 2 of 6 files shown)


zigaLuksic:
Please adjust the MR to target the develop branch, not the master one.

dajiaohuang:
Updated the PR base branch from master to develop as requested.

